PR #25, opened 2021-11-22
Aes capture

kralka:


scaaml/capture/acqktp.py
```diff
@@ -0,0 +1,226 @@
+from __future__ import division
+from __future__ import absolute_import
+
+from collections import OrderedDict
```
jmichelp, line 4: This seems to be a leftover from previous iterations of my code. This import isn't used anymore (we should consider having pylint or pyflakes and yapf workflows to check PRs)

scaaml/aes_forward.py
```diff
@@ -0,0 +1,35 @@
...
+        0xcd, 0x0c, 0x13, 0xec, 0x5f, 0x97, 0x44, 0x17, 0xc4, 0xa7, 0x7e, 0x3d, 0x64, 0x5d, 0x19, 0x73,
+        0x60, 0x81, 0x4f, 0xdc, 0x22, 0x2a, 0x90, 0x88, 0x46, 0xee, 0xb8, 0x14, 0xde, 0x5e, 0x0b, 0xdb,
+        0xe0, 0x32, 0x3a, 0x0a, 0x49, 0x06, 0x24, 0x5c, 0xc2, 0xd3, 0xac, 0x62, 0x91, 0x95, 0xe4, 0x79,
+        0xe7, 0xc8, 0x37, 0x6d, 0x8d, 0xd5, 0x4e, 0xa9, 0x6c, 0x56, 0xf4, 0xea, 0x65, 0x7a, 0xae, 0x08,
+        0xba, 0x78, 0x25, 0x2e, 0x1c, 0xa6, 0xb4, 0xc6, 0xe8, 0xdd, 0x74, 0x1f, 0x4b, 0xbd, 0x8b, 0x8a,
+        0x70, 0x3e, 0xb5, 0x66, 0x48, 0x03, 0xf6, 0x0e, 0x61, 0x35, 0x57, 0xb9, 0x86, 0xc1, 0x1d, 0x9e,
+        0xe1, 0xf8, 0x98, 0x11, 0x69, 0xd9, 0x8e, 0x94, 0x9b, 0x1e, 0x87, 0xe9, 0xce, 0x55, 0x28, 0xdf,
+        0x8c, 0xa1, 0x89, 0x0d, 0xbf, 0xe6, 0x42, 0x68, 0x41, 0x99, 0x2d, 0x0f, 0xb0, 0x54, 0xbb, 0x16,
+    ]
+
```
jmichelp, line 26: I don't remember how the training code works, but considering the attack points are described as strings in a JSON file and are also strings in a `dict` in the capture scripts, I guess it could make sense to have the name of the static methods match the text value we expect to use, and have a helper function:
```python
@classmethod
def get_attack_point(cls, name: str, **kwargs: bytearray) -> bytearray:
    function = getattr(cls, name, None)
    if not function or not callable(function):
        raise ValueError(f"{name} is not a defined attack point")
    return function(**kwargs)
```

WDYT?

scaaml/capture/aes_capture.py
```diff
@@ -0,0 +1,206 @@
+from pathlib import Path
+
+from scaaml.capture.aes_capture_util import *
+from scaaml.capture.acqktp import AcqKeyTextPattern_Scaaml as ktp_scaaml
+from scaaml.io import Dataset
+from scaaml.io.resume_kti import ResumeKTI
+
+
+class AESCaptureRunner:
```
jmichelp, line 9: According to the parameters, this class is very specific to captures done using exclusively Chipwhisperer hardware (scope + UART + clock generator). This is not the case for SCAAML datasets where the clock of the chip is different from the clock of the scope.

This should instead consider that we pass 4 objects in constructor, each implementing a defined API (and in specific cases, these can be the same object instance indeed):
* a scope object (chipwhisperer, picoscope, etc.)
* a control object (turning on/off the chip, resetting the chip, etc.) -- can be `None` in case we don't have a way to control the target board
* a communication object (chipwhisperer UART, external UART, etc.) -- this is typically the `target` parameter in Chipwhisperer API
* a crypto algorithm, providing the attack points and maybe some basic information about it (e.g. key length, etc.)

scaaml/capture/aes_capture.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(
+        self,
+        firmware_hash:
+        str = '18ba5773bbd766d5e6f76a6f1ebb689570ebed679232e22c8cf9ddbd140196d1',
```
jmichelp, line 13: I don't see the point about having a hardcoded sha256.
A few other default value don't really make sense here neither.

scaaml/capture/aes_capture.py
```diff
@@ -0,0 +1,206 @@
...
+        purpose: str = "train",
+        keys: int = 3072,
+        plaintexts: int = 256,
+        repetitions: int = 1,
+        samples: int = 7000,
+        offset: int = 0,
+        clock: int = 7372800,
+        gain: int = 45,
+        sample_rate: str = "clkgen_x4",
+        capture_info={},
```
jmichelp, line 29: Don't put mutable default values: they create a hidden class variable which can lead to unwanted side effects between object instances

scaaml/capture/aes_capture_util.py
```diff
@@ -0,0 +1,225 @@
...
+      cwscope: The chipwhisperer scope to use.
+      target: The chipwhisperer target to use.
+      resume_kti: An iterator of (key, text) pairs to capture traces with.
+      dataset: The dataset being created (for writing the actual traces).
+      examples_per_shard: How many captures are in a single shard.
+      plaintexts: How many plaintexts per a key value should be generated.
+      repetitions: How many traces should be captured with a single
+        (key, text) pair.
+      chip_id: Unique id of the piece of hardware.
+      purpose: Purpose of this dataset, in ['train', 'holdout'].
```
jmichelp, line 158: Then the typing could be more explicit:
```python
from typing import Literal

def capture_dataset(..., purpose: Literal['train', 'holdout']):
...
```

scaaml/capture/aes_capture_util.py
```diff
@@ -0,0 +1,225 @@
...
+                assert trace.textin == cur_pt
+                dataset.write_example(attack_points={
+                    'sub_bytes_in':
+                    AES_SBOX.sb_in(key=cur_key, text=cur_pt),
+                    'sub_bytes_out':
+                    AES_SBOX.sb_out(key=cur_key, text=cur_pt),
+                    'key':
+                    cur_key,
+                },
+                                      measurement={
```
jmichelp, line 203: weird indentation here.

scaaml/capture/aes_capture_util.py
```diff
@@ -0,0 +1,225 @@
...
+                                          "trace": trace.wave,
+                                      })
+                break  # from capturing the trace (while True)
+        # Increment the enumeration of the traces.
+        trace_number += 1
+    # Make sure we close the last shard
+    dataset.close_shard()
```
jmichelp, line 210: capturing may raise all sorts of exceptions (hardware failing, etc.) especially when it takes several hundreds hours to capture.
Don't we want something like a `contextmanager` to handle the resources nicely and make sure we close all the handles?
For example, when using Chipwhisperer scope, any failure at disconnecting/closing the resources (`cwscope.dis()` and `target.dis()`) will result in the hardware lock not being released. For there, one needs to physically replug the USB cable to access again the hardware. 

All functions/classes interacting with hardware should make sure that an uncaught exception will leave the hardware in a stable state where we can simply relaunch/resume the capture without physical intervention.

scaaml/capture/acqktp.py
kralka: I did a local pylint + yapf, a proper workflow is still a TODO.

scaaml/capture/aes_capture.py
kralka: Good point. I made it into a parameter.

As for other default values. I am happy to change those. Can you please elaborate which do not make sense and why?

scaaml/capture/aes_capture_util.py
kralka: yapf did this, I modified the code so it looks better.

scaaml/capture/aes_capture.py
kralka: I will try to remember this.

scaaml/capture/aes_capture_util.py
kralka: Done.

scaaml/capture/aes/scope.py
```diff
@@ -0,0 +1,80 @@
...
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        """Safely close all resources.
+
+        Args:
+          exc_type: None if no exception, otherwise the exception type.
+          exc_value: None if no exception, otherwise the exception value.
+          exc_tb: None if no exception, otherwise the traceback.
+        """
+        # Disconnect the scope if needed
+        self._scope.dis()
```
jmichelp, line 71: Is there a reason to disconnect, connect back to disable `target_pwr` (which belongs to `SControl`) to disconnect again?
I know Chipwhisperer doesn't make it easy as the scope is multifunction. `scope.io` belongs to `Scontrol`, `scope.adc`, `scope.clock`, `scope.gain` and `scope.trigger` belong to `SScope` and `scope.target` is `SCommunication`.

`scope.clock` is somehow also related to `SControl` but mainly because of the line `scope.io.hs2 = "clkgen"` which outputs the scope clock to the chip.

scaaml/capture/aes/crypto_alg.py
```diff
@@ -0,0 +1,215 @@
...
+            progress_filename=self._full_progress_filename)
+        # Set in get_stabilization_kti.
+        self._stabilization_ktp = None
+
+    @staticmethod
+    def attack_points_info():
+        """Returns the attack points info."""
+        # Number of classes used in ML. We are predicting a single byte => 256.
+        max_val = 256
+        return {
```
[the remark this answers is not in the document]
kralka: Done. Yes, the plan was to introduce base classes after implementing asymmetric crypto.

scaaml/capture/aes/scope.py
kralka: Re: disconnect and connect back was copied from old code... I removed that.

Otherwise done.

kralka:
@jmichelp Please take another look. I have done the changes in separate commits in the hope it will be easier for review. I intent to squash those before merge.

scaaml/capture/aes/capture_runner.py
```diff
@@ -0,0 +1,159 @@
...
+        except (Warning, OSError):
+            pass
+        # Stabilize the capture
+        for _ in range(10):
+            self.capture_trace(plaintext=cur_pt, key=cur_key)
+
+    def capture(self):
+        """Start (or resume) and finish the capture."""
+        for crypto_alg in self._crypto_algorithms:
+            self._stabilize_capture(crypto_alg=crypto_alg)
```
jmichelp, line 114: I think the stabilization can be factored out. We need to do it after a new firmware is flashed, after a power-cycle or things like this. But between datasets, this should be fine.

scaaml/capture/aes/aes_capture.py
```diff
@@ -0,0 +1,181 @@
...
+        add_crypto_alg(split='test',
+                       keys=test_keys,
+                       plaintexts=test_plaintexts,
+                       repetitions=repetitions)
+    if train_keys:
+        add_crypto_alg(split='train',
+                       keys=train_keys,
+                       plaintexts=train_plaintexts,
+                       repetitions=repetitions)
+    if holdout_keys:
```
jmichelp, line 156: We don't want to capture `holdout` with the same chip.
Valid combinations should be `train`, `train` and `test`, `test` (in case we already have `train`) or `holdout`

jmichelp (commented):
Please remove all ECC files. They don't belong here.

scaaml/capture/aes/crypto_alg.py
```diff
@@ -0,0 +1,205 @@
...
+          progress_filename: Filename to save progress (using resume_kti).
+          full_kt_filename: The file to save key-text pairs into. This should
+            be in the same directory as the whole dataset.
+          full_progress_filename: The file to save progress into. This should
+            be in the same directory as the whole dataset.
+        """
+        # TODO(karelkral): Automatically check scaaml.io.utils.sha256sum(firmware_file).
+        # When changing the following assert update also
+        # SCryptoAlgorithm._dataset
+        assert purpose in ['train', 'test', 'holdout']
```
jmichelp, line 48: Considering how often this is used, it might me worth having it declared in a `constants.py` file.
Not blocking for this PR though. Just something to keep in mind for further code cleaning.

scaaml/io/dataset.py
```diff
@@ -70,13 +80,18 @@ def __init__(
...
         self.capture_info = capture_info
         self.measurements_info = measurements_info
         self.attack_points_info = attack_points_info
 
         if not self.firmware_sha256:
             raise ValueError("Firmware hash is required")
+        if not self.firmware_url:
+            raise ValueError("Firmware URL is required")
 
         self.slug = "%s_%s_%s_v%s_%s" % (shortname, algorithm, architecture,
```
jmichelp, line 96: Nit: This could actually be turned into a f-string

jmichelp (approved):

scaaml/capture/aes/crypto_alg.py
kralka: Noted in issue #38 (same below)

Issue #31, opened 2021-12-16
Improve dataset meta information


ebursztein:
- Add Firmware url
- Add Licence  CC BY 4.0 > https://creativecommons.org/licenses/by/4.0/
- paper_url = ""
- scaaml_version  > implement a check that the Dataset lib >= scaaml_version in the dataset.

PR #33, opened 2021-12-22
Improve ds metainfo

kralka:
Fix issue #31 

This branch is on top of improve_ds_metainfo.

I am not sure if the last commit (making firmware_url) should be included, as it breaks loading of older datasets (those without firmware_url). I cannot see any solution that would be automatic and not overly complicated.

### DIFF
--- a/scaaml/utils.py
+++ b/scaaml/utils.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 "Utils common to various SCAAML components"
+from typing import List
+
 from tqdm.auto import tqdm
 import tensorflow as tf
 import numpy as np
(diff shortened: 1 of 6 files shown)


scaaml/utils.py
```diff
@@ -23,6 +25,22 @@
 from random import randint
 from termcolor import cprint, colored
 
+import scaaml
+
+
+def comparable_version(version: str = scaaml.__version__) -> List[int]:
```
jmichelp, line 31: You could [semver](https://pypi.org/project/semver/)

ebursztein (approved):

PR #35, opened 2022-01-05
Improve dataset metainformation

kralka:
Add the following metainformation:
  scaaml_version
  firmware_url
  paper_url
  licence

Fix issue #31


invernizzi:
Thanks Karel! LGTM. Only thing I spotted: since license/licence are valid spellings, maybe print out an alert if you find the unsupported version in the config (I'm partial to the American spelling, but I'm ok with either, as long as we're consistent)

kralka:
@invernizzi PTAL Thank you for the review and patience. I have implemented the spelling check. I left the 'licence' spelling to match issue #31

invernizzi:
Thank you, Karel!

Issue #36, opened 2022-01-06
Python3.7 compatibility


kralka:
Remove use of typing.Literal

PR #37, opened 2022-01-06
Remove use of typing.Literal

kralka:
Fix issue #36 Python3.7 compatibility.

### DIFF
--- a/scaaml/io/dataset.py
+++ b/scaaml/io/dataset.py
@@ -5,7 +5,7 @@ import json
 import os
 from collections import defaultdict
 from time import time
-from typing import Dict, List, Union, Literal
+from typing import Dict, List, Union
 from pathlib import Path
 import pprint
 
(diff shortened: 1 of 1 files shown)


scaaml/io/dataset.py
```diff
@@ -477,7 +477,7 @@ def summary(dataset_path):
 
     @staticmethod
     def inspect(dataset_path,
-                split: Literal['train', 'test', 'holdout'],
+                split: str,  # typing.Literal['train', 'test', 'holdout'],
```
jmichelp, line 480: Maybe we can be a bit smarter here:
```python
# Literal was introduced in Python 3.8
try:
  from typing import Literal
  SPLIT_TYPE = Literal['train', 'test', 'holdout']
except ImportError:
  SPLIT_TYPE = str
...
def inspect(dataset_path, split: SPLIT_TYPE, ...):
```

Up to you to decide.
Pros:
- we don't need to think about code changes when we'll drop py37 support

Cons:
- might be a bit less readable
- requires to disable some lint checks I presume

kralka: I would prefer not using this approach or at least doing this in a separate issue (I have created Issue #38 for various code improvements).

I plan to leave Literal in the code which is not run on the py3.7 computer (AES capture).

PR #43, opened 2022-01-17
Allow SScope to have unused keyword arguments

kralka:
The SScope is expected to be initialized using capture_info (to have all
information present in capture_info). But capture_info might contain
more information which is not needed to initialize SScope. This commit
allows SScope to ignore the rest of the capture_info.

### DIFF
--- a/scaaml/capture/aes/scope.py
+++ b/scaaml/capture/aes/scope.py
@@ -5,7 +5,7 @@ import chipwhisperer as cw
 class SScope:
     """Scope context manager."""
     def __init__(self, gain: int, samples: int, offset: int, clock: int,
-                 sample_rate: str):
+                 sample_rate: str, **_):
         """Create scope context.
 
         Args:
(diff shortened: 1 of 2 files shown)


kralka:
Part of issue #42

scaaml/capture/aes/scope.py
```diff
@@ -14,6 +14,21 @@ def __init__(self, gain: int, samples: int, offset: int, clock: int,
           offset: Number of samples to wait before starting recording data.
           clock: CLKGEN output frequency (in Hz).
           sample_rate: Clock source for cw.ClockSettings.adc_src.
+          _: SScope is expected to by initialized with capture_info which may
```
jmichelp, line 17: typo: to be

scaaml/capture/aes/scope.py
```diff
@@ -14,6 +14,21 @@ def __init__(self, gain: int, samples: int, offset: int, clock: int,
...
+            contain more keys. Thus we can ignore the rest of keyword
```
jmichelp, line 18: I don't find `more keys` easy to understand in this context. Either we need to add that `capture_info` is a `dict` or use `keywords`, like you did on the next sentence. WDYT?

jmichelp (approved):

scaaml/capture/aes/scope.py
jmichelp: Much better. Thanks

PR #44, opened 2022-01-19
Refactor scaaml capture into abstract classes

kralka:
Implement abstract classes to be shared with capture of other
cryptographic algorithms.

Part of issue #42

### DIFF
--- a/scaaml/capture/aes/communication.py
+++ b/scaaml/capture/aes/communication.py
@@ -30,8 +30,9 @@ class CWCommunication(AbstractSCommunication):
     def __enter__(self):
         """Initialize target."""
         assert self._target is None  # Do not allow nested with.
-        # the scope is there because of communication with the target (it communicated using single usb) the new cw uses a separate uart (import serial, the package is pyserial)
-        # 
+        # The scope is there because of communication with the target (it
+        # communicated using single USB endpoint). Since CW 5.5 firmware
+        # release it uses a separate USB UART.
         self._target = cw.target(self._scope, cw.targets.SimpleSerial)
         self._target.protver = self._protver
         self._scope = None
--- a/scaaml/capture/aes/communication.py
+++ b/scaaml/capture/aes/communication.py
@@ -15,10 +15,14 @@
 """The target in cw."""
 import chipwhisperer as cw
 
+from scaaml.capture.communication import AbstractSCommunication
 
-class SCommunication:
+
+class CWCommunication(AbstractSCommunication):
     """target in cw"""
     def __init__(self, scope):
+        """Initialize the communication object."""
+        super().__init__()
         self._scope = scope
         self._target = None
         self._protver = '1.1'
@@ -26,6 +30,8 @@ class SCommunication:
     def __enter__(self):
         """Initialize target."""
         assert self._target is None  # Do not allow nested with.
+        # the scope is there because of communication with the target (it communicated using single usb) the new cw uses a separate uart (import serial, the package is pyserial)
+        # 
         self._target = cw.target(self._scope, cw.targets.SimpleSerial)
         self._target.protver = self._protver
         self._scope = None
--- a/scaaml/capture/ascad/ascad_capture.py
+++ b/scaaml/capture/ascad/ascad_capture.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Google LLC
+# Copyright 2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -13,12 +13,13 @@
 # limitations under the License.
 
 """Capture script for easier manipulation."""
+
 from scaaml.aes_forward import AESSBOX
-from scaaml.capture.scope import CWScope
 from scaaml.capture.aes.aes_capture_contexts import capture_aes_dataset
+from scaaml.capture.scope import PicoScope
 
 
-def capture_aes_scald_stm32f4_mbedtls(
+def capture_ascad_default_parameters(
         firmware_sha256: str,
         crypto_implementation=AESSBOX,
         algorithm: str = 'simpleserial-aes',
@@ -31,33 +32,35 @@ def capture_aes_scald_stm32f4_mbedtls(
         examples_per_shard: int = 64,
         measurements_info=None,
         repetitions: int = 1,
-        chip_id: int = 164019,
+        chip_id: int = 620510,  # STM32F415 Training
         train_keys: int = 4 * 1024,
         train_plaintexts: int = 256,
         holdout_keys: int = 0 * 1024,
         holdout_plaintexts: int = 1) -> None:
-    """SCALD STM32F4 MBEDTLS specific defaults for capture_aes_dataset."""
-    architecture: str = 'CW308_STM32F4'
-    implementation: str = 'MBEDTLS'
-    shortname: str = 'aes_scald'
-    description: str = 'SCALD AES'
+    """SCAAML STM32F415 ASCAD specific defaults for capture_ascad_dataset."""
+    architecture: str = 'CW308_STM32F415'
+    implementation: str = 'ASCAD'
+    shortname: str = 'stm32f415_hwaes'
+    description: str = 'SCAAML HWAES'
 
     # Parameters for the scope
-    gain: int = 45
-    samples: int = 7000
+    samples: int = 160_000
+    sample_rate: float = 2.5e9  # Hz
     offset: int = 0
-    clock: int = 7372800
-    sample_rate: str = 'clkgen_x4'
+    trace_probe_range: float = 0.5  # V
+    trigger_range: float = 5.0  # V
+    trigger_level: float = 1.9  # V
     capture_info = {
-        'gain': gain,
         'samples': samples,
-        'offset': offset,
-        'clock': clock,
+        'trigger_level': trigger_level,
+        'trigger_range': trigger_range,
         'sample_rate': sample_rate,
+        'offset': offset,
+        'trace_probe_range': trace_probe_range,
     }
 
     capture_aes_dataset(
-        scope_class=CWScope,
+        scope_class=PicoScope,
         firmware_sha256=firmware_sha256,
         architecture=architecture,
         implementation=implementation,
--- a/scaaml/capture/capture_runner.py
+++ b/scaaml/capture/capture_runner.py
@@ -32,7 +32,7 @@ class AbstractCaptureRunner(ABC):
                  control: AbstractSControl, scope: AbstractSScope,
                  dataset: Dataset) -> None:
         """Holds all information needed to capture a dataset (using the method
-        capture).
+        `capture`).
 
         Args:
           crypto_algorithms: Provide attack points and information about the
--- a/scaaml/capture/capture_runner.py
+++ b/scaaml/capture/capture_runner.py
@@ -1,0 +1,155 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""CaptureRunner runs the capture."""
+from abc import ABC, abstractmethod
+from typing import Dict, List, Tuple
+from tqdm.auto import tqdm
+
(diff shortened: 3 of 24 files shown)


scaaml/capture/scope.py
```diff
@@ -0,0 +1,54 @@
...
+from abc import ABC, abstractmethod
+
+
+class AbstractSScope(ABC):
+    """Scope context manager."""
+    def __init__(self, gain: int, samples: int, offset: int, clock: int, **_):
+        """Create scope context.
+
+        Args:
+          gain: Gain of the scope.
```
jmichelp, line 12: `gain` and `clock` are specific to Chipwhisperer. A more traditional scope wouldn't have such controls.
Normally you deal with `samplerate` (how many samples per second you want to capture), `timebase` expressed in seconds (or ms, us, ns, etc.) per division (a typical scope screen has 10 divisions and the timebase is the horizontal resolution), `vertical scale` expressed in volts per division (a typical scope has 10 divisions).

So `samples` would be `timebase * 10 * samplerate`, `gain` is the vertical scale but with Chipwhisperer it's the amplification factor in dB, and `clock` is related to the `samplerate` as the Chipwhisperer can acquire either at `1x` or `4x` of this clock speed.

So if you really want an abstract class, it shouldn't be tainted by Chipwhisperer oddities IMO :)

scaaml/capture/aes/communication.py
```diff
@@ -1,10 +1,14 @@
 """The target in cw."""
 import chipwhisperer as cw
 
+from scaaml.capture.communication import AbstractSCommunication
 
-class SCommunication:
+
+class SCommunication(AbstractSCommunication):
```
jmichelp, line 7: With the protocol version and the mention of a scope, this means that this communication implementation is meant to use a Chipwhisperer.
For other captures where OpenADC (the scope embedded in the Chipwhisperer boards) cannot be used, I usually really on an external UART.

Maybe this class should be named `ChipwhispererComm` or `CWCommunication`?

scaaml/capture/aes/control.py
```diff
@@ -2,36 +2,25 @@
 
 from time import sleep
 
+from scaaml.capture.control import AbstractSControl
 
-class SControl:
+
+class SControl(AbstractSControl):
```
jmichelp, line 8: Like the scope class, this one is also specific to Chipwhisperer hardware.
In general case, when we don't use the Chipwhisperer capture board, we can't reset or power cycle the chip.
So I suggest to use this refactoring to rename this class to make it more obvious that it's related to Chipwhisperer only

scaaml/capture/capture_runner.py
```diff
@@ -0,0 +1,141 @@
...
+            _, _ = self.get_attack_points_and_measurement(
+                crypto_alg=crypto_alg, crypto_input=crypto_input)
+        except (Warning, OSError, AssertionError):
+            pass
+        # Stabilize the capture
+        for _ in range(10):
+            _, _ = self.get_attack_points_and_measurement(
+                crypto_alg=crypto_alg, crypto_input=crypto_input)
+
+    def capture(self):
```
jmichelp, line 94: I know we already discussed it but from the capture point of view, the fact that ML requires 2 datasets for training should be transparent.
If we capture `training` the tooling should take care of capturing (and eventually resuming in case of interruption) `train` and `test` subsets all at once. The geometry of `test` should be inferred from the requested one for `train`.

scaaml/capture/aes/control.py
```diff
@@ -49,32 +38,15 @@ def reset(self,
...
         sleep(sleep_after)
 
+    def initialize(self):
+        """Initialize the target."""
+        self._scope_io.tio1 = self._tio1
+        self._scope_io.tio2 = self._tio2
+        self._scope_io.hs2 = self._hs2
+
     def turn_on(self):
         """Turn on the chip."""
```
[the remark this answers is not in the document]
kralka: Good point

scaaml/capture/capture_runner.py
kralka: I thought this was going to be another PR. But I am happy to do it here :)

scaaml/capture/ascad/ascad_capture.py
```diff
@@ -0,0 +1,274 @@
...
+                          holdout_plaintexts=holdout_plaintexts)
+
+
+def load_firmware():
+    """Load firmware."""
+    FIRMWARE = "chipwhisperer-firmware/simpleserial-aes/simpleserial-aes_CW308-STM32F4_HWAES.hex"
+    cwscope = cw.scope()
+    cwscope.default_setup()
+    try:
+        cw.program_target(cwscope, cw.programmers.STM32FProgrammer, FIRMWARE)
```
jmichelp, line 270: Loading firmware through the Chipwhisperer has been quite unreliable in the past, not to mention that it only supports a handful of targets.
Consider moving to the code I wrote to program the chips using an external programmer. It should support all the target chips that can be purchased with the Chipwhisperer and provide a consistant API and Notebook widgets :)

scaaml/capture/ascad/ascad_capture.py
```diff
@@ -0,0 +1,274 @@
...
+        measurements_info=None,
+        repetitions: int = 1,
+        train_keys: int = 3 * 1024,
+        train_plaintexts: int = 256,
+        holdout_keys: int = 0 * 1024,
+        holdout_plaintexts: int = 1) -> None:
+    """Capture or continue capturing the dataset.
+
+    Args:
+      firmware_sha256: Hash of the used firmware binary.
```
jmichelp, line 40: nit: IntelHEX file. Because we also have ELF files but these are not reproducible from a compilation to another, yielding different hashes while the HEX files are stable.

scaaml/capture/ascad/scope.py
```diff
@@ -0,0 +1,131 @@
+"""Context manager for the scope."""
```
jmichelp, line 1: Something I only noticed now: all the files owned by Google must start with the license header with the correct years mentioned.
Not blocking this PR as I think other files outside of this PR already have been merged but please do a PR to fix it.

scaaml/capture/ascad/picoscope.py
```diff
@@ -0,0 +1,447 @@
+#!/usr/bin/python
```
jmichelp, line 1: Please add a comment to tell that this files originally comes from Chipwhisperer project. It's not Google code. I think I needed to do some modifications on it or support for it has been dropped when they moved to py3

scaaml/capture/ascad/scope.py
```diff
@@ -0,0 +1,131 @@
...
+
+        Returns: self
+        """
+        assert self._scope is None  # Do not allow nested with.
+        picoscope = PicoScope6404()
+        picoscope.con()
+        picoscope.trace.range = self._trace_probe_range
+        picoscope.trace.channel = "A"
+        picoscope.trace.coupling = "AC"
+        picoscope.trace.probe_attenuation = "1:1"  # golden T
```
jmichelp, line 48: nit: s/golden T/BNC-SMA pigtail/

scaaml/capture/ascad/scope.py
```diff
@@ -0,0 +1,131 @@
...
+        #picoscope.trace.probe_attenuation = "1:10"  # Alligator + hook
```
jmichelp, line 49: nit: oscilloscope passive probe

scaaml/capture/scope.py
```diff
@@ -0,0 +1,48 @@
...
+from abc import ABC, abstractmethod
+
+
+class AbstractSScope(ABC):
+    """Scope context manager."""
+    def __init__(self, samples: int, offset: int, **_):
+        """Create scope context.
+
+        Args:
+          samples: How many points to sample (length of the capture).
```
jmichelp, line 12: nit: How many samples to capture

scaaml/capture/scope.py
```diff
@@ -0,0 +1,48 @@
...
+          offset: Number of samples to wait before starting recording data.
```
jmichelp, line 13: nit: to wait after trigger event occurred before starting ...

scaaml/capture/ascad/ascad_capture.py
```diff
@@ -0,0 +1,274 @@
...
+      architecture: Architecture of the used chip.
+      implementation: Name of the implementation that was used.
+      shortname: Short name of the dataset being captured.
+      description: Short description for the scaaml.io.Dataset.
+      capture_info: Used as parameters of scaaml.io.SScope. Should contain:
+        samples: int: Number of data points in a single capture = length of the
+          trace (see cw documentation).
+        offset: int: After how many samples to start recording (see cw
+          documentation).
+        sample_rate: str: Clock source for cw.ClockSettings.adc_src (see cw
```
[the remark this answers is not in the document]
kralka: That is true, moreover the type was wrong. But we are using this in aes: https://github.com/google/scaaml/blob/master/scaaml/capture/aes/scope.py#L16 and the cw documentation calls it ClockSettings.adc_src scope.clock.adc_src (https://chipwhisperer.readthedocs.io/en/latest/api.html#chipwhisperer.capture.scopes._OpenADCInterface.OpenADC.ClockSettings.adc_src)
The clock source for the ADC module. (but that is a string)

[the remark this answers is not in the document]
kralka: I see, then I would suggest leaving it as after 8ecabe989e0fe19b06a26a9db6cb386711440178 WDYT?

scaaml/capture/scope/cw_scope.py
```diff
@@ -43,29 +46,23 @@ def __init__(self, gain: int, samples: int, offset: int, clock: int,
...
-        Args:
-          gain: Gain of the scope.
-          samples: How many points to sample (length of the capture).
-          offset: Number of samples to wait before starting recording data.
-          clock: CLKGEN output frequency (in Hz).
-          sample_rate: Clock source for cw.ClockSettings.adc_src.
+        Returns: self
         """
         assert self._scope is None  # Do not allow nested with.
         cwscope = cw.scope()
```
jmichelp, line 68: I think here you should directly assign `self._scope`.
Because I'm afraid that if an exception occurs while updating the settings, the context manager may call `__exit__()` and `self._scope` would still be `None`.

scaaml/capture/scope/default_cw_scope.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __enter__(self):
+        """Create scope context.
+
+        Returns: self
+        """
+        assert self._scope is None  # Do not allow nested with.
+
+        # Open cw scope with default settings.
+        cwscope = cw.scope()
```
jmichelp, line 38: same as above, it's probably safer to assign it to `self._scope` as early as possible in case exceptions arise while setting up the scope

scaaml/capture/scope/picoscope.py
```diff
@@ -0,0 +1,106 @@
...
+        picoscope.trigger.coupling = "DC"
+        picoscope.trigger.probe_attenuation = "1:1"
+        picoscope.trigger.range = self._trigger_range
+
+        # Number of samples settings.
+        picoscope.sample_rate = self._sample_rate
+        picoscope.sample_length = self._samples
+        picoscope.sample_offset = self._offset
+
+        self._scope = picoscope
```
jmichelp, line 89: Same as above with earlier assign

scaaml/capture/scope/scope_base.py
```diff
@@ -0,0 +1,63 @@
...
+class AbstractSScope(ABC):
+    """Scope context manager."""
+    def __init__(self, samples: int, offset: int, **_):
+        """Create scope context.
+
+        Args:
+          samples: How samples to capture (length of the capture).
+          offset: Number of samples to wait after trigger event occurred before
+            starting recording data.
+          _: CWScope is expected to be initialized using the capture_info
```
jmichelp, line 29: Suggestion: make it more generic.
For example: derived classes may require additional parameters to properly configure the scope.

scaaml/capture/aes/communication.py
```diff
@@ -15,17 +15,23 @@
...
+        """Initialize the communication object."""
+        super().__init__()
         self._scope = scope
         self._target = None
         self._protver = '1.1'
 
     def __enter__(self):
         """Initialize target."""
         assert self._target is None  # Do not allow nested with.
+        # the scope is there because of communication with the target (it communicated using single usb) the new cw uses a separate uart (import serial, the package is pyserial)
```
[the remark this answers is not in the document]
kralka: No, I forgot. I need to setup workflows...

scaaml/capture/scope/scope_base.py
jmichelp: It still sounds a bit weird to document specificities of Chipwhisperer scope inside an abstract class. That's what my comment is about.

kralka: I see, thank you. Then I suggest to remove it as we do not use this parameter anywhere (in the last commit). I added one more commit adding missing argument documentation of PicoScope.

jmichelp (approved):

Issue #49, opened 2022-02-11
Randomness checks


kralka:
Implement checks that the attack points and traces seem random enough (and in case of traces correlated between different splits).

Framework where it is easy to add other tests is needed. Preferably single pass over the data (parsing seems slow, but not measured yet).

kralka:
Trace tests: It might be a good idea to be able to manualy inspect a small set of traces and report those that are most different for further inspection.

kralka:
More advanced: https://en.wikipedia.org/wiki/Diehard_tests

PR #55, opened 2022-03-07
Print of all warnings and statistics

kralka:
Pretty print of attack point check warnings and trace statistics.

Issue #49

### DIFF
--- a/scaaml/stats/print_stats.py
+++ b/scaaml/stats/print_stats.py
@@ -74,12 +74,11 @@ class PrintStats:
                         for shard in dataset.shards_list[split])
         prod = list(product(all_splits, all_groups, all_parts))
         for split, group, part in tqdm(prod):
-            print(f'split: {split} group: {group} part: {part}')
             example_iterator = ExampleIterator(dataset_path=dataset_path,
                                                split=split,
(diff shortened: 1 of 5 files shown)


scaaml/stats/print_stats.py
```diff
@@ -0,0 +1,146 @@
...
+        all_parts = set(shard['part'] for split in Dataset.SPLITS
+                        for shard in dataset.shards_list[split])
+        prod = list(product(all_splits, all_groups, all_parts))
+        for split, group, part in tqdm(prod):
+            print(f'split: {split} group: {group} part: {part}')
+            example_iterator = ExampleIterator(dataset_path=dataset_path,
+                                               split=split,
+                                               group=group,
+                                               part=part)
+            for example in tqdm(example_iterator):
```
invernizzi, line 82: Having a tqdm loop within a tqdm loop will likely cause artifacts to be left on the screen (that is: do you see a bunch of progress bars?).

If you change the innermost loop to: `for example in tqdm(example_iterator, leave=False):` the issue should go away.

invernizzi (approved):

invernizzi (approved):

Issue #38, opened 2022-01-06
Improve code quality


kralka:
Ideas:

- Set up automation of tests and code quality checks
- Add docstrings
- Improve formatting (yapf)
- Fix typing (mypy)
- Work in linter advice (pylint)
- Run spell checker

kralka:
Fix typing idea: Use Literal type hint for split (bevare of Python3.7 compatibility) https://github.com/google/scaaml/pull/37#discussion_r779524015

kralka:
General refactoring: Turn old formatting into f-strings.

jmichelp:
FYI
With PRs #63 and #64 being merged, I now enforced yapf to be green to merge any new PR.

kralka:
It could be nice having dataclasses for parameters that we use everywhere together. For instance: https://github.com/google/scaaml/pull/75#discussion_r911128880

PR #61, opened 2022-05-25
Enable pylint and yapf workflow jobs

kralka:
Part of issue #38

### DIFF
--- a/.github/python_matcher.json
+++ b/.github/python_matcher.json
@@ -1,0 +1,40 @@
+{
+    "problemMatcher": [
+        {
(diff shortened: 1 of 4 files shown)


kralka:
Workflow from https://github.com/google/OpenSK

Changes:
python.yml Removed ignore of third_party, ubuntu-22.04, removed tockloader
.pylintrc 4 spaces indentation
run_pylint.sh Remove ignore of third_party

pylint fails with InconsistentMroError because of tqdm.auto

jmichelp (approved):

PR #63, opened 2022-05-27
Fix code formatting and associated workflow

jmichelp:
Part of issue #38 

Yapf code style was using 2-spaces indentation while scaaml was using 4. So I moved to the google style which matches what we currently have.

Python files were modified by running the following command and checking the diff manually:
```shell
yapf --style=google --recursive --in-place .
```


kralka (approved):
LGTM thank you

PR #64, opened 2022-05-27
Split workflows and adds initial support for pytest and mypy

jmichelp:
Part of issue #38 

Because we're using google codestyle in yapf, pylint configuration
file has been updated to the matching one in Google public style guide:
https://github.com/google/styleguide/blob/gh-pages/pyguide.md

Rationale behind splitting workflows is the granularity of branch protection which is at the yaml file level.
It's also not so bad for maintenance considering it's a python project.

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -11,10 +11,10 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.7
+      - name: Set up Python 3.9
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7
+          python-version: 3.9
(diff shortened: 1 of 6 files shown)


jmichelp:
We'll probably need to develop the associated matchers for both mypy and pytest to bring the annotations back into the review tool. But that can be done later. At least we already have the information in the execution trace.

kralka (commented):
Thank you

kralka (approved):

kralka (dismissed):

jmichelp:
Ok, looks good now. We'll see if the cache is working (for this pytest and/or mypy must pass successfully) first. And I'll revisit the workflows if it doesn't work as planned.

kralka (approved):

PR #66, opened 2022-06-01
More easy pylint fixes

kralka:
Fix most of pylint problems (17 left)

### DIFF
--- a/scaaml/capture/scope/ps6424e.py
+++ b/scaaml/capture/scope/ps6424e.py
@@ -332,7 +332,7 @@ class Pico6424E(ChipWhispererCommonInterface):
         """
         # Handle too large sample_rate
         if sample_rate > 5e9:
-            raise ValueError('This scope support at most 5GHz sample_rate.')
+            raise ValueError("This scope support at most 5GHz sample_rate.")
 
         # From PicoScope API manual:
         # https://www.picotech.com/download/manuals/picoscope-6000-series-a-api-programmers-guide.pdf
@@ -422,7 +422,7 @@ class Pico6424E(ChipWhispererCommonInterface):
         # Check if this scope is connected.
         if self.connectStatus is False:
             raise Exception(
-                f'Scope {self._name} is not connected. Connect it first.')
+                f"Scope {self._name} is not connected. Connect it first.")
 
         # Run the capture block
         timeIndisposedMs = ctypes.c_double(0)
@@ -498,7 +498,7 @@ class Pico6424E(ChipWhispererCommonInterface):
         Returns: np array representing the last captured trace.
         """
         if as_int:
-            msg = 'Returning trace as integers is not implemented.'
+            msg = "Returning trace as integers is not implemented."
             raise NotImplemented(msg)
 
         return np.array(self._buffers[0][:], dtype=np.float32)
@@ -513,7 +513,7 @@ class Pico6424E(ChipWhispererCommonInterface):
         """Setup channels, buffers, and trigger."""
         if self.ps_handle.value <= 0:
             # No opened PicoScope handle
-            msg = f'Scope handle is {self.ps_handle.value}'
+            msg = f"Scope handle is {self.ps_handle.value}"
             assert not self.connectStatus, msg
             return
 
@@ -579,12 +579,12 @@ class Pico6424E(ChipWhispererCommonInterface):
         # If we use hardware downsampling, use averaging.
         if self.DOWNSAMPLING_RATIO > 1:
             self._downsampling_mode = picoEnum.PICO_RATIO_MODE[
-                'PICO_RATIO_MODE_AVERAGE']
-            # 'PICO_RATIO_MODE_DECIMATE' could also be an option.
+                "PICO_RATIO_MODE_AVERAGE"]
+            # "PICO_RATIO_MODE_DECIMATE" could also be an option.
         else:
             self._downsampling_mode = picoEnum.PICO_RATIO_MODE[
-                'PICO_RATIO_MODE_RAW']
-        data_type = picoEnum.PICO_DATA_TYPE['PICO_INT16_T']
+                "PICO_RATIO_MODE_RAW"]
+        data_type = picoEnum.PICO_DATA_TYPE["PICO_INT16_T"]
         waveform = 0
         # Set trace buffer
         action = picoEnum.PICO_ACTION["PICO_CLEAR_ALL"] | picoEnum.PICO_ACTION[
@@ -654,9 +654,9 @@ class Pico6424E(ChipWhispererCommonInterface):
     def __repr__(self) -> str:
         """Return device name, connected status and dict representation as
         multi-line string."""
-        connected = 'Connected' if self.connectStatus else 'Not connected'
+        connected = "Connected" if self.connectStatus else "Not connected"
         dict_repr = util.dict_to_str(self._dict_repr())
-        return f'{self._name} device {connected}\n{dict_repr}'
+        return f"{self._name} device {connected}\n{dict_repr}"
 
     def __str__(self) -> str:
         return self.__repr__()
--- a/scaaml/capture/scope/ps6424e.py
+++ b/scaaml/capture/scope/ps6424e.py
@@ -5,7 +5,6 @@ from __future__ import absolute_import
 
 import ctypes
 from decimal import Decimal, ROUND_HALF_DOWN
-import time
 import traceback
 from typing import OrderedDict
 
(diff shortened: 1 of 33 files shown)


scaaml/capture/scope/ps6424e.py
```diff
@@ -498,12 +506,13 @@ def get_last_trace(self, as_int: bool = False) -> np.ndarray:
...
-            msg = 'Returning trace as integers is not implemented.'
-            raise NotImplemented(msg)
+            msg = "Returning trace as integers is not implemented."
+            raise NotImplementedError(msg)
 
         return np.array(self._buffers[0][:], dtype=np.float32)
 
-    getLastTrace = util.camel_case_deprecated(get_last_trace)
+    # Legacy naming.
+    getLastTrace = util.camel_case_deprecated(get_last_trace)  # pylint: disable=C0103
```
jmichelp, line 515: Same as above

scaaml/capture/scope/ps6424e.py
```diff
@@ -540,7 +549,7 @@ def _set_channels(self):
             ps.ps6000aSetChannelOn(
                 self.ps_handle,  # handle
                 self.trace.ps_api_channel,  # channel
-                self.trace._coupling,  # coupling
+                self.trace._coupling,  # coupling  # pylint: disable=W0212
```
jmichelp, line 552: Instead of disabling the warning, what about adding a `coupling_picoapi` or `coupling_internal` get property?

scaaml/io/dataset.py
```diff
@@ -623,23 +625,23 @@ def _check_disjoint_keys(self, pbar, key_ap: str, deep_check: bool = True):
         Raises: ValueError if some key from train is present in test.
         """
         seen_keys = set()
-        for i in range(len(self.shards_list['test'])):
+        for i in range(len(self.shards_list["test"])):
```
jmichelp, line 628: I thought `range(len(...))` was a discouraged pattern nowadays.
If it is, maybe replace with:
```python
for i, _ in enumarate(self.shards_list[Dataset.TEST_SPLIT]):
```

And in all cases, the text constant should be replaced for consistency

scaaml/io/dataset.py
```diff
@@ -623,23 +625,23 @@ def _check_disjoint_keys(self, pbar, key_ap: str, deep_check: bool = True):
...
                                 dpath=self.path,
-                                train_shards=self.shards_list['train'],
+                                train_shards=self.shards_list["train"],
                                 pbar=pbar,
                                 examples_per_shard=self.examples_per_shard,
                                 key_ap=key_ap)
         else:
             Dataset._shallow_check(seen_keys=seen_keys,
-                                   train_shards=self.shards_list['train'],
+                                   train_shards=self.shards_list["train"],
```
jmichelp, line 644: Same again

scaaml/io/dataset.py
```diff
@@ -675,41 +677,41 @@ def _check_shard_metadata(shard_info: Dict, dataset_path: Path) -> None:
...
-        if size != shard_info['size']:
-            raise ValueError(f'Wrong size, got: {size}, expected: '
-                             f'{shard_info["size"]}, in shard: {shard_info}')
+        size = os.stat(dataset_path / shard_info["path"]).st_size
+        if size != shard_info["size"]:
+            raise ValueError(f"Wrong size, got: {size}, expected: "
+                             f"{shard_info['size']}, in shard: {shard_info}")
         # Check chip_id is non-negative integer
-        chip_id = shard_info['chip_id']
+        chip_id = shard_info["chip_id"]
```
jmichelp, line 711: Reviewing this code, I'm thinking that all the text constants that are dictionary keys should all be declared as a constants in a class instead of repeating them all over the place.
Otherwise any change in the schema will be horrible to do and one typo somewhere can lead to unexpected behaviour

scaaml/io/dataset.py
```diff
@@ -1058,21 +1059,21 @@ def cleanup_shards(dataset_path):
...
         """
         fpath = Dataset._get_config_path(dataset_path)
 
         # Save the old config.
-        save_path = Path(f'{str(fpath)}.sav.{time()}.json')
+        save_path = Path(f"{str(fpath)}.sav.{time()}.json")
         assert not save_path.exists()  # Do not overwrite.
         save_path.write_text(fpath.read_text())
-        cprint("Saving old config to %s" % save_path, 'cyan')
+        cprint(f"Saving old config to {save_path}", "cyan")
```
jmichelp, line 1070: Maybe also we may want to make sure that color printing give a consistant UX throughout our code and instead of writing the color, on each `cprint()` call, we could have a bunch of helpers that wrap them (e.g. `cprint_info()`, `cprint_warning()`, `cprint_error()`, etc.).

scaaml/plot.py
```diff
@@ -70,13 +71,13 @@ def plot_comparaison(traces, labels, title):
...
         label = labels[idx]
         if label == "SNR":
-            color = '#8E24AA'
-        elif 'activation' in label.lower():
-            color = '#64DD17'
-        elif 'grad' in label.lower():
-            color = '#26A69A'
-        elif 'scald' in label.lower():
-            color = '#03A9F4'
+            color = "#8E24AA"
```
jmichelp, line 74: I wonder how hard this would be to use a lookup `dict` here instead.
But at least we could factor out `label.lower`:
```python
label = labels[idx].lower()
if label == "snr":
...
```

scaaml/utils.py
```diff
@@ -21,52 +21,52 @@
...
     for e in lst:
         h.append(pretty_hex(e))
     cprint(prefix + " ".join(h), color)
 
 
 def get_model_stub(attack_point, attack_byte, config):
-    return '%s-%s-%s-v%s-ap_%s-byte_%s-len_%s' % (
-        config['device'], config['algorithm'], config['model'],
-        config['version'], attack_point, attack_byte, config['max_trace_len'])
+    return f"{config['device']}-{config['algorithm']}-{config['model']}-" \
```
jmichelp, line 55: parentheses instead backslashes

scaaml_intro/train.py
```diff
@@ -11,39 +11,43 @@
...
 
 
 def train_model(config):
     tf_cap_memory()
-    TRAIN_GLOB = "datasets/%s/train/*" % config['algorithm']
-    TEST_GLOB = "datasets/%s/test/*" % config['algorithm']
-    TEST_SHARDS = 256
-    NUM_TRACES_PER_TEST_SHARDS = 16
-    BATCH_SIZE = config['batch_size'] * get_num_gpu()
+    TRAIN_GLOB = f"datasets/{config['algorithm']}/train/*"  # pylint: disable=C0103
```
jmichelp, line 34: We should rename these instead of disabling the linter.
These are not file-level nor class-level constants and should therefore not be capitalized

tools/run_pylint.sh
```diff
@@ -18,7 +18,8 @@ SUCCESS=0
 # Ensure we are at the project root directory
 cd $(readlink -f $(dirname $0))/..
 
-for file in `find . -type f -name '*.py'`
+# For .py files which are not in tests or notebook checkpoints
+for file in `find . ! \( -path "./tests/*" -o -path "*/.ipynb_checkpoints/*" \) -type f -name '*.py'`
```
jmichelp, line 22: I'm worried that find may behave differently on OS X and on Linux here.
How about replacing this with:
```sh
for file in `git ls-files --full-name --deduplicate --exclude-standard '*.py'`
```

This only considers the files that are already part of the repository and not already ignored by our `.gitignore` file.
The drawback is that it can't ignore `tests/` directory as it's part of the repo.
What's the rationale behind ignoring it?

scaaml/capture/scope/ps6424e.py
```diff
@@ -644,8 +653,8 @@ def sample_offset(self, val):
 
     def _dict_repr(self):
         ret = OrderedDict()
-        ret["trace"] = self.trace._dict_repr()
-        ret["trigger"] = self.trigger._dict_repr()
+        ret["trace"] = self.trace._dict_repr()  # pylint: disable=W0212
```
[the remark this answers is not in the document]
kralka: Agree, but on the other hand it matches cw. So I am not sure if we prefer to be closer to the original counterpart or have a nicer API https://github.com/newaetech/chipwhisperer/blob/599139c4aea9c28603198bc372315af83d889867/software/chipwhisperer/capture/scopes/OpenADC.py#L803

tools/run_pylint.sh
kralka: Rationale of not pylinting tests was to allow more gradual linter coverage (not doing everything at once) and the fact that tests seemed like a good candidate of what to do in a later iteration. That being said I like git ls-files better.

jmichelp (approved):

PR #68, opened 2022-06-02
Assume pylint 2.14 or above is installed

jmichelp:


tools/run_pylint.sh
```diff
@@ -13,19 +13,8 @@
...
-for file in `find . -type f -name '*.py'`
-do
-  # Output header for our custom matcher on Github workflow
-  echo "PYLINT:${file}"
-  if ! pylint --rcfile=.pylintrc --score=n "$file"
-  then
-    SUCCESS=1
-  fi
-done
+pylint *.py scaaml scaaml_intro tests tools
```
kralka, line 19: Is there a reason to name the directories explicitly?

pylint *.py .

kralka (approved):

tools/run_pylint.sh
kralka: SGTM, thank you for explaining :)

PR #71, opened 2022-06-03
Pylint fixes3

kralka:


tests/test_aes_forward.py
```diff
@@ -33,26 +33,32 @@ def test_bs_out():
     # yapf: disable
```
jmichelp, line 33: Then there should be no point at disabling yapf for these areas (note that the `yapf: enable` at the end should then also be removed)

tests/test_aes_forward.py
```diff
@@ -67,30 +73,38 @@ def test_inverse():
...
```
jmichelp, line 73: same here

tests/test_aes_forward.py
```diff
@@ -111,13 +125,16 @@ def test_attack_points():
     """test getting different attack points using AESSBOX.get_attack_point"""
     # yapf: disable
```
jmichelp, line 126: and here

update_dependency_licenses.py
```diff
@@ -29,12 +30,14 @@ def GetDependencies():
     """Extracts dependencies from setup.py"""
     dependencies = []
 
-    def setup(**kwargs):
+    # pylint set_up fails, setup is the name in setuptools, pylint has an
+    # extempt on setUp.
```
jmichelp, line 34: typo?

jmichelp (approved):
So much green in the workflows! 🍾

PR #76, opened 2022-06-08
Fix pytest

kralka:
Part of issue #38

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -20,6 +20,12 @@ jobs:
         run: |
           python -m pip install --upgrade pip setuptools wheel
           pip install pytest
+          # Install PS drivers
+          wget -O - https://labs.picotech.com/Release.gpg.key | sudo apt-key add -
+          sudo bash -c 'echo "deb https://labs.picotech.com/rc/picoscope7/debian/ picoscope main" >/etc/apt/sources.list.d/picoscope7.list'
+          sudo apt-get update
+          sudo apt-get install libps6000a
+          sudo apt-get install libpswrappers
       - name: Installing package
         run:
           pip install -r requirements.txt


scaaml/capture/scope/scope_template.py
```diff
@@ -0,0 +1,53 @@
...
+chipwhisperer API."""
+
+from abc import abstractmethod
+
+import numpy as np
+
+from chipwhisperer.capture.api.cwcommon import ChipWhispererCommonInterface
+
+
+class ScopeTemplate(ChipWhispererCommonInterface):
```
jmichelp, line 24: Don't derive from `ChipwhispererCommonInterface`. The name is a bit misleading but if you look at the methods it has, they aren't related to the scope at all (firmware updates, etc.).

We should simply get our class, then have a `ChipwhispererScope` that derives from `ScopeTemplate` and `OpenADC`

kralka: I am a bit lost here. I don't understand where a `ChipwhispererScope` that derives from `ScopeTemplate` and `OpenADC` comes into.

On the cw side: `OpenADC` derives from `ChipwhispererCommonInterface` [[1]](https://github.com/newaetech/chipwhisperer/blob/1a31817b2de5ac3608631ed08c09809a2d980c03/software/chipwhisperer/capture/scopes/OpenADC.py#L54). And `cw.capture_trace` argument scope [[2]](https://github.com/newaetech/chipwhisperer/blob/1a31817b2de5ac3608631ed08c09809a2d980c03/software/chipwhisperer/__init__.py#L373) has type `ScopeTypes = Union[OpenADC, CWNano]` (this is why I was afraid that mypy would consider passing any other type to be an error).

On scaaml side: We have only context managers (`CWScope, DefaultCWScope, PicoScope`) which have `.scope` property. And a single scope class acting as a `ScopeTypes` (`Pico6424E`).

scaaml/capture/scope/scope_template.py
```diff
@@ -39,7 +46,7 @@ def arm(self):
         """Setup scope to begin capture when triggered."""
 
     @abstractmethod
-    def capture(self):
+    def capture(self, poll_done: bool = False) -> bool:
```
jmichelp, line 50: Same thing here regarding prototype. I don't see what `poll_done` would bring. Removing it from the function signature and putting the default value in `Chipwhisperer.capture()` should do the trick for us. It's not a value we change.
If later we need extra arguments to support a specific scope, we can revisit it by adding `**kwargs` to the function.

Please document return value.

scaaml/capture/scope/scope_template.py
```diff
@@ -28,7 +28,8 @@ def __init__(self):
         """Initialize the base."""
 
     @abstractmethod
-    def con(self,
+    def con(
```
jmichelp, line 31: I think that for the abstraction, `def con(self, device_identifier: Optional[str] = None, **kwargs)` should be an appropriate signature.
`device_identifier` being a way to disambiguate which scope we want it case multiple scopes from the same model are plugged id and detected. Default behavior when set to `None` is to connect to the first one found.
You can decide to either put `**kwargs` as part of the signature or simply remove it because we don't need it at the moment for the scopes we support. When we'll need it it will be easy to add it anyway.

This means that we'll have to implement `def con()` for `ChipwhispererScope` so that it passes on the correct default values to `super().con()`.

kralka: With your signature I am getting `error: Definition of "con" in base class "ScopeTemplate" is incompatible with definition in base class "OpenADC"`

scaaml/capture/scope/scope_template.py
kralka: Same error with incompatible definition. `ChipwhispererScope` inherits both `OpenADC` and `ScopeTemplate` thus the base classes must have same signatures. Am I not understanding something?

scaaml/capture/scope/scope_template.py
kralka: I think Python supports changing the signature when inheriting (`Pico6424E.con `signature is different from `ScopeTemplate.con` signature), but when doing multiple inheritance the signatures of superclasses must match. Other thing is that cw.capture_trace expects the argument scope to be of type Union[OpenADC, CWNano].

I do not understand the role of `ChipwhispererScope` in your last code snippet. We have a `CWScope` and `DefaultCWScope` context managers (auto-con/dis the `OpenADC`) and then we have a `Pico6424E` which plays a role of an object that behaves as an `OpenADC` (and is passed to cw API instead of `OpenADC`), then we have a context-manager with `PicoScope.scope` is `Pico6424E`. Finally we have reimplemented `ScopeTemplate` which is an abstract class capturing the cw scope interface.

Over all I agree with your comments in the sense of not polluting our API with unnecessary compatibility with CW. The thing is that I am not sure how to do that and still have a working typechecking. And I also agree that typechecking is far from perfect in the current PR (my idea was fix the workflows and then slowly improve the rest).

kralka:
I suggest to resolve the inheritance question in another PR (issue: https://github.com/google/scaaml/issues/79)

invernizzi (approved):
Thank you for all the changes, Karel!

PR #87, opened 2022-07-07
[REVERT] Check all files for spelling errors

kralka:
Fix most of the typos so they do not block new PRs. Some of the typos were not fixed (parts of API and notebooks) see https://github.com/google/scaaml/issues/90

Part of https://github.com/google/scaaml/issues/38


project-words.txt
```diff
@@ -1,19 +1,96 @@
...
+batchnorm
+batchnormalization
+bibtex
+booktitle
+Bursztein
+burszteindc
+burzteindc
 bytelist
 chipwhisperer
+chipwispher
```
jmichelp, line 20: This one seems like a typo somewhere in the doc/code rather than a word we should add.

project-words.txt
```diff
@@ -1,19 +1,96 @@
...
+clkgen
+cmap
+colorbar
+comparaison
```
jmichelp, line 24: Same here. That's the French word for comparison :)

tests/capture/scope/test_ps6424e.py
```diff
@@ -22,9 +22,11 @@
 def test_get_timebase_too_large():
     """Test sample rate to timebase conversion for too large sampling rate."""
     # Test too large
-    with pytest.raises(ValueError) as verror:
+    with pytest.raises(ValueError) as value_error:
         timebase = Pico6424E._get_timebase(1 + 5e9)
-    assert "This scope support at most 5GHz sample_rate." == str(verror.value)
+    expected_msg: str = "This scope support at most 5GHz sample_rate."
```
jmichelp, line 27: supports

setup.py
```diff
@@ -40,7 +40,6 @@
         "matplotlib",
         "Pillow",
         "tensorflow>=2.2.0",
-        "future-fstrings",
```
jmichelp, line 43: I think this was required to support an early python3 version (maybe 3.6?).
Isn't there a place in `setup.py` to declare the minimum supported python3 version?

project-words.txt
kralka: Agree, but I wanted to be careful with fixing typos that change API during this PR (it is large and otherwise full of rather save typo fixes). See issue https://github.com/google/scaaml/issues/90

Part of function name at https://github.com/google/scaaml/blob/master/scaaml/utils.py#L99

That being said I am happy to fix it in this PR, but I thought it would be easier to review this separately.

project-words.txt
kralka: Similar as above, part of function name at https://github.com/google/scaaml/blob/b01f2bb685938082b423a3f112aaa934c8da10f1/scaaml/plot.py#L68

My French vocabulary is growing :)

jmichelp (approved):

PR #91, opened 2022-07-07
Implement our own spell check action

kralka:
Do not rely on another projects action


kralka:
@jmichelp 

There are two solutions with their pros and cons.

1. Run cspell on all changed files. But only the first 10 typos get annotated (10errors, 10warnings, 10notes, total at most 50 per job). Thus there is a potential of discovering typos that have not been annotated before (those were still present in the output).
2. Use a matrix and run a job for every changed file separately. The code is more complicated and some jobs can get canceled if others fail -> also not good, but at least the canceled jobs are visible. Also if a file is ignored, then cspell fails (it fails when run on zero files). This can be fixed by running it on two files (the real changed file and a file we know is there and not ignored).

Another con is that the gui shows only typos in the changed parts of changed files...

All and all I would suggest the easy solution with cspell on all changed files. Or even better fixing all typos (maybe gradually by file type) and running cspell on all files.

WDYT?

.github/workflows/spellcheck.yml
```diff
@@ -8,8 +8,46 @@ on:
...
+        uses: actions/checkout@v2
+        with:
+          # We need all history to list all changed files.
+          fetch-depth: 0
+      - name: Generate matrix
+        id: generate-matrix
+        run: |
+          # List all files that changed between master and head of the
+          # current branch. Encode the output as json.
+          FILES="$(git diff --name-only origin/master..${{ github.event.after }} | jq -cnR '[inputs | select(length>0)]')"
```
jmichelp, line 27: Don't if this impacts your workflow or not but here, you're storing in an env variable which is limited in terms of capacity. If this is an issue, you should output into a temporary file instead to work around that limitation.

jmichelp (dismissed):
Running cspell on all changed files seems good.

Just curious: where does the limitation of 10 reported errors comes from? Is it a cspell thing, a matcher or a github workflow limit?

kralka:
> Running cspell on all changed files seems good.
> 
> Just curious: where does the limitation of 10 reported errors comes from? Is it a cspell thing, a matcher or a github workflow limit?

https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md#limitations

The limit should not be a problem (after a while).

jmichelp (approved):

PR #92, opened 2022-07-11
Pylint with Python3.10

kralka:
Pylint also with Python3.10. Other workflows (mypy, pytest...) are still run against Python3.9. Suggested at https://github.com/google/scaaml/pull/87#discussion_r917855371

@jmichelp Should we also bump Python version for other workflows?


kralka:
I suggest to bump Python version in workflows to 3.10 and when 3.11 is out to add it to the pylint matrix.

jmichelp:
Sounds good.
Going back in versions wouldn't make sense.

Be careful that tensorflow sometimes needs time to fully test a new version of python interpreter. Adding 3.11 too early could result in a missing package.

jmichelp (approved):

PR #93, opened 2022-07-20
Make resume_kti more general

### LINKED ISSUE #48: Allow resume_kti to save other type of data than just key, plaintext
The key, plaintext pair may not be what is needed in different cryptographic primitives (and using plaintext for something else seems error prone). Even more values might be needed.

kralka:
Allow any set of parameters (key, plaintext, mask...) to be saved using
resume_kti.

Fixes #48

### DIFF
--- a/scaaml/capture/aes/crypto_alg.py
+++ b/scaaml/capture/aes/crypto_alg.py
@@ -81,8 +81,10 @@ class SCryptoAlgorithm(AbstractSCryptoAlgorithm):
             texts.append(list(text))
         # Does not overwrite existing keys-texts file.
         self._kti = resume_kti.create_resume_kti(
-            keys=np.array(keys_list, dtype=np.uint8),
-            texts=np.array(texts, dtype=np.uint8),
+            parameters={
+                "keys": np.array(keys_list, dtype=np.uint8),
+                "texts": np.array(texts, dtype=np.uint8),
+            },
             shard_length=np.uint64(self._examples_per_shard),
             kt_filename=self._full_kt_filename,
             progress_filename=self._full_progress_filename)
@@ -119,7 +121,7 @@ class SCryptoAlgorithm(AbstractSCryptoAlgorithm):
                 # AcqKeyTextPatternScaaml.new_pair raises StopIteration itself.
                 kt_pair = self._ktp.new_pair()
                 # Allow the same iteration as using resume_kti.
-                return list(kt_pair[0]), list(kt_pair[1])
+                return {"keys": list(kt_pair[0]), "texts": list(kt_pair[1])}
 
         self._stabilization_ktp = StabilizationIterator(self._get_new_ktp())
         return self._stabilization_ktp
--- a/scaaml/capture/aes/crypto_alg.py
+++ b/scaaml/capture/aes/crypto_alg.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """AES specific SCryptoAlgorithm."""
+from collections import namedtuple
 from typing import Iterator, Optional
 import numpy as np
 
@@ -113,6 +114,8 @@ class SCryptoAlgorithm(AbstractSCryptoAlgorithm):
 
             def __init__(self, ktp):
                 self._ktp = ktp
+                self._element_class = namedtuple("EncryptionParameters",
+                                                 ["keys", "texts"])
 
             def __iter__(self):
                 return self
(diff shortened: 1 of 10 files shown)


scaaml/io/resume_kti.py
```diff
@@ -11,15 +11,17 @@
...
 Autosaves after shard length iterations.
 
 Typical usage example:
 
   # Save all key-text pairs ahead before the experiment starts.
   # If kt_filename exists this does nothing.
-  create_resume_kti(keys, texts, shard_length=shard_length,
+  create_resume_kti(parameters={"keys": keys, "texts": texts},
+                    shard_length=shard_length,
                     kt_filename="key_text_pairs.txt",
```
jmichelp, line 25: now they're not pairs anymore :)

scaaml/capture/aes/crypto_alg.py
```diff
@@ -119,7 +121,7 @@ def __next__(self):
                 # AcqKeyTextPatternScaaml.new_pair raises StopIteration itself.
                 kt_pair = self._ktp.new_pair()
                 # Allow the same iteration as using resume_kti.
-                return list(kt_pair[0]), list(kt_pair[1])
+                return {"keys": list(kt_pair[0]), "texts": list(kt_pair[1])}
```
jmichelp, line 124: maybe that's just personal preference, but a namedtuple here would look nicer than a dict for the return value of the iterator

scaaml/io/resume_kti.py
jmichelp, line 25: Technically it's not a pair anymore if the dict has more than 2 keys. How about renaming these to parameters_tuples.txt and progress_tuples.txt?

scaaml/io/resume_kti.py
```diff
@@ -165,15 +237,18 @@ def __next__(self):
         """
         if self._index % self._shard_length == 0:
             self._save_progress()
-        if self._index >= len(self._keys):
+        if self._index >= get_any_value_len(self._parameters):
             raise StopIteration
         self._index += 1
-        return self._keys[self._index - 1], self._texts[self._index - 1]
+        return {
```
jmichelp, line 243: Same here, I think a namedtuple would look nicer. You can always turn the namedtuple into a dict using the `._asdict()` method or get a tuple with the field names using `._fields` property

tests/io/test_resume_kti.py
```diff
@@ -28,8 +28,8 @@ def test_save_and_load(tmp_path):
     # Check that the files do not exist now
     assert not os.path.isfile(tmp_path / KT_FILENAME)
     assert not os.path.isfile(tmp_path / PROGRESS_FILENAME)
-    resume_kti = create_resume_kti(keys=KEYS,
-                                   texts=TEXTS,
+    parameters = {"keys": KEYS, "texts": TEXTS}
+    resume_kti = create_resume_kti(parameters=parameters,
```
jmichelp, line 32: Maybe we should modify one of the tests (or add a new one) to include a mask with a different length and make sure everything still works.

scaaml/io/resume_kti.py
```diff
@@ -232,22 +238,23 @@ def __next__(self):
...
-        }
-
-    def __len__(self):
-        """How many pairs are there in total (including the initial_index
-        skipped pairs)."""
+        return self._element_class(
+            **{
+                parameter_name: self._parameters[parameter_name][self._index -
+                                                                 1]
+                for parameter_name in self._parameters  # pylint: disable=C0206
```
jmichelp, line 252: This comprehension isn't great to read, especially because of lines 250 and 251 being split.

scaaml/io/resume_kti.py
kralka: I tried to remove pair. But there is still the naming of keys, plaintexts in Dataset (where plaintexts is the number of other parameters used with each key value).

scaaml/capture/aes/crypto_alg.py
kralka: We get nicer unpacking. Agreed.

scaaml/io/resume_kti.py
```diff
@@ -160,20 +238,25 @@ def __next__(self):
...
         self._index += 1
-        return self._keys[self._index - 1], self._texts[self._index - 1]
 
-    def __len__(self):
-        """How many pairs are there in total (including the initial_index
-        skipped pairs)."""
-        return len(self._keys)
+        # Construct the namedtuple to return.
+        element_class_parameters = {
+            parameter_name: self._parameters[parameter_name][self._index - 1]
```
jmichelp, line 251:
```python
element_class_parameters = {
    parameter_name: val[self._index - 1]
    for parameter_name, val in self._parameters.items()
}
```

wouldn't the above allow us to remove the pylint disable?

kralka:
PTAL

jmichelp (approved):

PR #99, opened 2022-08-30
Move example iterator

kralka:


packages/scaaml_dataset/tests/io/test_example_iterator.py
```diff
@@ -18,7 +18,7 @@
 import numpy as np
 
 from scaaml.io import Dataset
-from scaaml.stats import ExampleIterator
+from scaaml_dataset.io import ExampleIterator
```
jmichelp, line 21: I'm not sure about best practice here.
But maybe we can consider scaaml-dataset as being a package belonging to the overall scaaml namespace.
Each package would contribute with its own files.

But maybe that's too complicated to setup when things aren't split across multiple repositories.

kralka: I didn't know about namespaces. So not sure about best practices at all.

Just to be sure in the end we want to have separate repositories, but instead of:

from scaaml_dataset.io import ExampleIterator

we would write:

from scaaml.dataset.io import ExampleIterator

The second one looks nicer, but I like that the first one directly gives you the package name. Also scaaml-dataset is not taken as a package name, no idea how naming works on pypi with namespaces.

.github/workflows/mypy.yml
```diff
@@ -22,7 +22,8 @@ jobs:
           pip install mypy
       - name: Installing package
         run:
-          pip install -r requirements.txt
+          pip install .
+          pip install packages/scaaml_dataset
```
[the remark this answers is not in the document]
kralka: Good catch, this should be left as was. The thing is that for now the packages are co-dependent scaaml.legacy.stats.print_stats imports scaaml.dataset.io.ExampleIterator and scaaml.dataset.io.ExampleIterator imports scaaml.legacy.io.Dataset. My solution while moving things around was to have the top-level setup.py script install also all the packages.

jmichelp (approved):

PR #110, opened 2022-12-05
Fix typing inside untyped functions

kralka:
Fix typing, only non-trivial change should be scaaml/io/shard.py (change of value of maximum from 0 to -infinity, as that makes more sense).

### DIFF
--- a/scaaml/intro/generator.py
+++ b/scaaml/intro/generator.py
@@ -14,6 +14,7 @@
 """Dataset creation and loading."""
 
 import tensorflow as tf
+from tensorflow import Tensor
 import numpy as np
 from tensorflow.keras.utils import to_categorical
 from termcolor import cprint
@@ -41,8 +42,8 @@ def create_dataset(file_pattern,
         raise ValueError(
             "invalid attack point. avail: key, sub_bytes_in, sub_bytes_out")
 
-    x = []
-    y = []
+    x_list: List = []
+    y_list: List = []
     pb = tqdm(total=num_shards, desc="loading shards")
     with tf.device("/cpu:0"):
         for idx, shard_fname in enumerate(shards):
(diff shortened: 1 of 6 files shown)


scaaml/intro/generator.py
```diff
@@ -57,14 +60,14 @@ def create_dataset(file_pattern,
...
-            y.append(y_shard)
+            x_list.append(x_shard)
+            y_list.append(y_shard)
             pb.update()
         pb.close()
         # Disable pylint warnings due to
         # https://github.com/PyCQA/pylint/issues/3613
-        x = tf.concat(x, axis=0)  # pylint: disable=E1120,E1123
-        y = tf.concat(y, axis=0)  # pylint: disable=E1120,E1123
+        x: Tensor = tf.concat(x_list, axis=0)  # pylint: disable=E1120,E1123
```
jmichelp, line 69: Do we still have to disable warnings or has it been fixed?

scaaml/io/dataset.py
```diff
@@ -162,7 +162,8 @@ def __init__(
 
         # current shard tracking
         self.shard_key: Optional[str] = None
-        self.prev_shard_key = None  # track key change for counting
+        self.prev_shard_key: Optional[
+            str] = None  # track key change for counting
```
jmichelp, line 166: For readability: move the comment above the line so that the code isn't split

scaaml/capture/scope/default_cw_scope.py
```diff
@@ -14,7 +14,10 @@
 """Context manager chipwhisperer scope that has default setting and is used to
 control."""
 
+from typing import Optional, Union
```
jmichelp, line 17: If it's unused, do we need to keep this line?

scaaml/capture/scope/cw_scope.py
```diff
@@ -13,9 +13,13 @@
 # limitations under the License.
 """Context manager for the scope."""
 
+from typing import Optional, Union
```
jmichelp, line 16: If it's unused, do we need to keep this line?

scaaml/io/dataset.py
kralka: Done

PR #111, opened 2022-12-05
All actions running node 16

kralka:
Fix warning, more info:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -10,7 +10,7 @@ jobs:
   mypy:
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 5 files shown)


kralka:
I can still see warnings connected to 'save-state' and 'set-output', but we do not call those directly. So let's hope we are fine... (GitHub is deprecating those start of the year 2023.) More info: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

jmichelp (approved):

PR #120, opened 2023-03-28
Fix inconsistency in the rank function.

kralka:
When all predictions are different (most frequent case) the optimistic parameter should not play a role in the result.


kralka:
Pylint addressed in https://github.com/google/scaaml/pull/120 waiting for that one.

invernizzi (approved):

PR #170, opened 2023-10-17
Do not capture test when not capturing train

wsxrdv:


scaaml/capture/aes/aes_capture_contexts.py
```diff
@@ -107,6 +107,9 @@ def capture_aes_dataset(
     # If we are capturing train also capture the test split. We want the SBOX
     # input to be balanced, thus we need at least 256 test_keys.
     test_keys = max(train_keys // 8, 256)  # 1/8 = 0.125
+    # Do not capture test when not capturing train.
+    if train_keys == 0:
```
jmichelp, line 111: Might be worth unit-testing this part

kralka:
Soon to be replaced by configuration based by https://github.com/google/scaaml/tree/main/scaaml/capture/input_generators

PR #179, opened 2023-12-04
Fix typehint of variable

kralka:
https://github.com/google/scaaml/actions/runs/7083392728


kralka:
I will add this to the original dependabot PR

PR #181, opened 2023-12-04
Allow AES capture using LeCroy oscilloscope

wsxrdv:


kralka:
Should be `._scope_communication` todo test real capture and fix

scaaml/capture/aes/aes_capture_contexts.py
```diff
@@ -258,6 +258,24 @@ def _capture(scope_class, capture_info: Dict[str, Any], chip_id: int,
...
+    # Capture using LeCroy.
+    if scope_class == LeCroy:
+        with LeCroy(**capture_info) as lecroy:
+            # Update capture info with oscilloscope details.
+            dataset.capture_info.update(lecroy.scope.get_identity_info())
+            dataset.write_config()
+
+            assert lecroy.scope is not None
+            with DefaultCWScope(capture_info.get("cw_scope_serial_number",
+                                                 None)) as default_cwscope:
```
jmichelp, line 270: I think `dict.get("xxx")` should be equivalent to `dict.get("xxx", None)`

scaaml/capture/scope/lecroy/lecroy.py
```diff
@@ -281,3 +281,19 @@ def __str__(self) -> str:
...
+
+    def get_identity_info(self) -> Dict[str, str]:
+        """Get information about the oscilloscope identity.
+
+        Returns: a dictionary containing the model, serial_number, and
+        firmware_level (version).
+        """
+        assert self._scope_communication
+        answer = self._scope_communication.query("*IDN?").rstrip()
+        lecroy, model, serial_number, firmware_level = answer.split(",")
```
jmichelp, line 293: nit: maybe `brand` instead of `lecroy`

scaaml/capture/aes/aes_capture_contexts.py
```diff
@@ -258,6 +258,24 @@ def _capture(scope_class, capture_info: Dict[str, Any], chip_id: int,
         generators.
       dataset (scaaml.io.Dataset): The dataset to save examples to.
     """
+    # Capture using LeCroy.
+    if scope_class == LeCroy:
+        with LeCroy(**capture_info) as lecroy:
```
jmichelp, line 263: With 3 different scopes, I wonder if code wouldn't look better by rewriting it the following way:
```python
with scope_class(**capture_info) as scope_object:
  scope_object.init(dataset=dataset)
  assert scope_object.scope
  with DefaultCWScope(capture_info.get("cw_scope_serial_number)) as default_cwscope:
    _control_communication_and_capture(
        ....
    )
```

And for each scope class you add the method:
```python
def init(self, **kwargs: Dict[Any, Any]) -> None:
  pass
```
Which for the Lecroy would then be:
```python
def init(self, dataset: scaaml.io.Dataset, **kwargs: Dict[Any, Any]) -> None:
  dataset.capture_info.update(self.scope.get_identity_info())
  dataset.write_config()

File: scaaml/capture/aes/aes_capture_contexts.py [hunk 4]
```diff
@@ -258,37 +258,20 @@ def _capture(scope_class, capture_info: Dict[str, Any], chip_id: int,
         generators.
       dataset (scaaml.io.Dataset): The dataset to save examples to.
     """
+    with scope_class(**capture_info) as scope_context:
+        # Save information ackquired from the scope
```
jmichelp, line 262: typo: acquired

jmichelp (approved):

Issue #192, opened 2024-01-25
Code coverage workflow


kralka:
Add a code coverage workflow (how well the code is unittested, new code comes with code coverage...)

kralka:
@LucaTomasko what do you think about the following? https://coverage.readthedocs.io/en/7.4.0/ combined with https://github.com/google/scaaml/blob/main/.github/workflows/pytest.yml

jmichelp:
As mentioned in #221 a problem matcher won't really work in this scenario.

Workflow is in place and running. Should we close this issue?

PR #195, opened 2024-01-26
Integrate coveralls

kralka:
Issue #192


.github/workflows/pytest.yml
```diff
@@ -8,6 +8,7 @@ on:
     types: [opened, synchronize, reopened]
 env:
   MIN_COVERAGE_REQUIRED: 70
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jmichelp, line 10: don't put the token in the env for the whole action.
Only add it to the env of the step that uploads to coveralls. We only need it here.

.github/workflows/pytest.yml
```diff
@@ -25,6 +25,7 @@ jobs:
         run: |
           pip install pytest
           pip install coverage
+          pip install coveralls
```
jmichelp, line 28: nit: merge all 3 `pip install`?

.github/workflows/pytest.yml
```diff
@@ -36,6 +33,6 @@ jobs:
       - name: Running unit tests
         run: |
           coverage run --omit="tests/**/*.py" -m pytest
-      - name: Report code coverage
+      - name: Upload results
         run: |
-          coverage report -i -m --fail-under=$MIN_COVERAGE_REQUIRED
+          GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} coveralls --service=github
```
jmichelp, line 38:
```suggestion
- name: Upload results
  run: coveralls --service=github
  env: 
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

jmichelp (approved):

PR #200, opened 2024-02-01
Implement repeat iterator

LucaTomasko:
Issue #196

### DIFF
--- a/scaaml/capture/input_generators/attack_point_iterator.py
+++ b/scaaml/capture/input_generators/attack_point_iterator.py
@@ -61,6 +61,7 @@ def _build_attack_points_iterator(configuration: Dict) -> AttackPointIterator:
         "constants": AttackPointIteratorConstants,
         "balanced_generator": AttackPointIteratorBalancedGenerator,
         "unrestricted_generator": AttackPointIteratorUnrestrictedGenerator,
+        "repeat": AttackPointIteratorRepeat,
     }
     operation = configuration["operation"]
     iterator_cls = supported_operations.get(operation)
@@ -152,3 +153,30 @@ class AttackPointIteratorUnrestrictedGenerator(AttackPointIterator):
 
     def get_generated_keys(self) -> List[str]:
         return [self._name]
+
+
+class AttackPointIteratorRepeat(AttackPointIterator):
+    """
+    Attack point iterator class that iterates 
+    over a configuration a repeated amount of times.
+    """
+
+    def __init__(self, operation: str, repetitions: int,
+                 configuration: Dict) -> None:
+        """Initialize the repeated iterate."""
+        assert "repeat" == operation
+        self._repetitions = repetitions
+        self._configuration_iterator = build_attack_points_iterator(
+            configuration)
+        self._len = repetitions * len(self._configuration_iterator)
+
+    def __len__(self) -> int:
+        return self._len
+
+    def __iter__(self):
+        return iter(
+            list(iter(self._configuration_iterator))
+            for repetitions in range(self._repetitions))
+
+    def get_generated_keys(self) -> List[str]:
+        return self._configuration_iterator.get_generated_keys()
--- a/tests/capture/input_generators/test_attack_point_iterator.py
+++ b/tests/capture/input_generators/test_attack_point_iterator.py
@@ -131,3 +131,44 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
     output = list(iter(build_attack_points_iterator(config)))
     assert len(output) == config["bunches"] * config["elements"]
     assert len(output) == len(build_attack_points_iterator(config))
+
+
+def test_attack_point_iterator_repeat():
+    config = {
+        "operation": "repeat",
+        "repetitions": 2,
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [1, 2, 3]
+        }
+    }
+    output = list(iter(build_attack_points_iterator(config)))
+    assert output == [[{
+        'key': 1
+    }, {
+        'key': 2
+    }, {
+        'key': 3
+    }], [{
+        'key': 1
+    }, {
+        'key': 2
+    }, {
+        'key': 3
+    }]]
+
+
+def test_attack_point_iterator_repeat_len():
+    config = {
+        "operation": "repeat",
+        "repetitions": 2,
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [1, 2, 3]
+        }
+    }
+    output = len(build_attack_points_iterator(config))
+    assert output == config["repetitions"] * len(
+        config["configuration"]["values"])


scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +153,30 @@ def __iter__(self):
...
+        self._configuration_iterator = build_attack_points_iterator(
+            configuration)
+        self._len = repetitions * len(self._configuration_iterator)
+
+    def __len__(self) -> int:
+        return self._len
+
+    def __iter__(self):
+        return iter(
+            list(iter(self._configuration_iterator))
```
jmichelp, line 178: Be careful, calling `list()` here will trigger pre-computing all the iterator values in-memory. Depending on the configuration passed as a parameter, that could crash

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -131,3 +131,44 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+    config = {
+        "operation": "repeat",
+        "repetitions": 2,
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [1, 2, 3]
+        }
+    }
+    output = len(build_attack_points_iterator(config))
```
kralka, line 172: `output_length = len(list(iter(build_attack_points_iterator(config))))`

A subtle bug. When a configuration should be returning:
```
{'key': 1}
{'key': 2}
{'key': 3}
```
Then repeat two times should be returning
```
{'key': 1}
{'key': 2}
{'key': 3}
{'key': 1}
{'key': 2}
{'key': 3}
```
Not the following
```
[{'key': 1}, {'key': 2}, {'key': 3}]
[{'key': 1}, {'key': 2}, {'key': 3}]
```

scaaml/capture/input_generators/attack_point_iterator.py
kralka:
```
def __iter__(self):
  return self

def __next__(self):
  for _ in range(self._repetitions):
    for attack_points in self._configuration_iterator:
      yeild attack_points
```

kralka: Please add the following test to test that the memory blow-up mentioned by JM does not happen:

```
def test_repeat_memory():
    config = {
        "operation": "repeat",
        "repetitions": 1_000_000_000_000,
        "configuration": {
            "operation": "constants",
            "name": "key",
            "values": [1, 2, 3]
        }
    }

    long_iterator = iter(build_attack_points_iterator(config))
    assert next(long_iterator) == { "key": 1 }
    assert next(long_iterator) == { "key": 2 }
    assert next(long_iterator) == { "key": 3 }
    assert next(long_iterator) == { "key": 1 }
    assert next(long_iterator) == { "key": 2 }
    assert next(long_iterator) == { "key": 3 }
    assert next(long_iterator) == { "key": 1 }
```

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +153,29 @@ def __iter__(self):
 
     def get_generated_keys(self) -> List[str]:
         return [self._name]
+
+
+class AttackPointIteratorRepeat(AttackPointIterator):
+    """
+    Attack point iterator class that iterates 
+    over a configuration a repeated amount of times.
```
kralka, line 161: over a configuration a `repeat` number of times.

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +153,29 @@ def __iter__(self):
...
+    """
+
+    def __init__(self, operation: str, repetitions: int,
+                 configuration: Dict) -> None:
+        """Initialize the repeated iterate."""
```
kralka, line 166: Documentation what are possible values of arguments. Especially for `repetitions` what happens when that is 0 or negative? Is there a way to repeat forever? Unittests missing. Check with the design doc.

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +153,29 @@ def __iter__(self):
...
+        self._repetitions = repetitions
+        self._configuration_iterator = build_attack_points_iterator(
+            configuration)
+        self._len = repetitions * len(self._configuration_iterator)
+
+    def __len__(self) -> int:
+        return self._len
+
+    def __iter__(self):
+        return iter(value for repetition in range(self._repetitions)
```
kralka, line 177: NIT Two forloops nested in iter expression are not the most readable (not clear which happens first). But ok with me if you are 100% sure it is correct.

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -131,3 +132,70 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+        "repetitions": 2,
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [1, 2, 3]
+        }
+    }
+    output = list(iter(build_attack_points_iterator(config)))
+    assert output == list(
+        iter(build_attack_points_iterator(config["configuration"]))) * 2
```
kralka, line 149: I like that this is configuration agnostic, but it could be even better `* config["repetitions"]`

kralka (commented):
Thank you.

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +153,35 @@ def __iter__(self):
...
+    Attack point iterator class that iterates 
+    over a configuration a repeat amount of times.
+    """
+
+    def __init__(self,
+                 operation: str,
+                 configuration: Dict,
+                 repetitions: int = 0) -> None:
+        """Initialize the repeated iterate if repetitions is not present
+          or set to below 0 it will do an infinite loop."""
```
kralka, line 169: Arguments documentation missing. See for instance: https://github.com/google/scaaml/blob/main/scaaml/capture/input_generators/input_generators.py#L44

Example use missing.

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +153,35 @@ def __iter__(self):
...
+        assert "repeat" == operation
+        if repetitions > 0:
+            self._repetitions = repetitions
```
kralka, line 172: There is no unittest for length in the infinite case. There is no unittest testing that there are at least some (say two or 3 elements being iterated when `repetitions = 0`).

kralka (commented):
Thank you, PTAL at the comments

scaaml/capture/input_generators/attack_point_iterator.py
LucaTomasko: But wouldn't they be in the configuration Dict?

kralka: The documentation is not in the configuration dictionary. Where can a user find documentation of what happens when they set `repetitions` to zero, negative number, positive number...? This is a non-trivial thing that is not documented anywhere. We could add that documentation to the whole `build_attack_points_iterator` function. But it needs to be somewhere.

LucaTomasko: It is in the init doc string

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +153,37 @@ def __iter__(self):
...
+class AttackPointIteratorRepeat(AttackPointIterator):
+    """
+    Attack point iterator class that iterates 
+    over a configuration a repeat amount of times.
+    """
+
+    def __init__(self,
+                 operation: str,
+                 configuration: Dict,
+                 repetitions: int = 0) -> None:
```
kralka, line 167: `0` should be "no iteration", `-1` should be infinite.

kralka: The iteration should default to "infinite" iteration, not to no iteration. At least according to the design doc. If you want to change it feel free to leave a comment or a suggestion.

scaaml/capture/input_generators/attack_point_iterator.py
kralka: Outdated comment: when `repetitions = -1`

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +154,42 @@ def __iter__(self):
...
+        """Initialize the repeated iterate if repetitions is not present
+          or set to 0 it will not repeat and if it set 
+          to a negative number it will do an infinite loop."""
```
kralka, line 171: Where can I find documentation of different values of the `repetitions` parameter?

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +154,42 @@ def __iter__(self):
...
+        assert "repeat" == operation
+        self._configuration_iterator = build_attack_points_iterator(
+            configuration)
+        if repetitions > 0:
+            self._repetitions = repetitions
+            self._len = repetitions * len(self._configuration_iterator)
+        elif repetitions == 0:
```
kralka, line 178: This is redundant (result is the same as if you had the first condition `if repetitions >= 0:` (ok here you save one call to `len`).

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -145,3 +146,113 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+    config = {
+        "operation": "repeat",
+        "repetitions": -1,
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [1, 2, 3]
+        }
+    }
+    with pytest.raises(AttributeError):
```
kralka, line 220: Technically not according to the design doc. I am ok with changing that. But

1) where is this behavior documented?

2) `AttributeError` is not what an end-user would expect.

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -145,3 +146,113 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+
+
+def test_attack_point_iterator_repeat():
+    config = {
+        "operation": "repeat",
+        "repetitions": 2,
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [1, 2, 3]
```
kralka, line 158: Some tests are missing. I would like to see tests for `repetitions in [-1, 0, 1, 2]` and for `len(values) in [0, 1, 2, 3]`. Especially the edge case of `repetitions = -1` and `values = []` is not covered (documented and the code does not produce the expected result, compare with `itertools.cycle([])`).

kralka (commented):
Thank you

tests/capture/input_generators/test_attack_point_iterator.py
LucaTomasko: What should it return then because we can't return infinite as a length

kralka: There are two solutions:

1) as the design doc suggests iterate at most `sys.maxsize` iterations.

2) cycle infinitely, but then there needs to be a change in the design doc & the error must be informative and documented.

LucaTomasko: So I have implemented infinite with the cycle function but for the len() function should I create a detailed custom exception?

kralka: That sounds good. And please document it in the docstring.

kralka: And also don't forget to suggest a change to the design doc.

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -43,6 +44,11 @@ def get_generated_keys(self) -> List[str]:
         """
 
 
+class LengthIsInfiniteException(Exception):
+    """This exception get's called when a number is infinite
+      and wants to be represented by __len__ function."""
```
kralka, line 49: """This exception is raised when the `__len__` function is called on an infinite iterator."""

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +159,43 @@ def __iter__(self):
...
+        """Initialize the repeated iterate if repetitions is not present
+          or set to an negative number it will do an infinite loop 
+          if it is 0 it will not repeat at all."""
+        assert "repeat" == operation
+        self._configuration_iterator = build_attack_points_iterator(
+            configuration)
+        if repetitions >= 0:
+            self._repetitions = repetitions
+            self._len = repetitions * len(self._configuration_iterator)
+
```
kralka, line 183: NIT extra blank line

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +159,43 @@ def __iter__(self):
...
+    """
+    Attack point iterator class that iterates 
+    over a configuration a repeat amount of times.
+    """
+
+    def __init__(self,
+                 operation: str,
+                 configuration: Dict,
+                 repetitions: int = -1) -> None:
+        """Initialize the repeated iterate if repetitions is not present
```
kralka, line 174: Much better, still no information about the possibility of raising an exception. Also not according to the style guide https://google.github.io/styleguide/pyguide.html#383-functions-and-methods (another comment had a linked example https://github.com/google/scaaml/blob/main/scaaml/capture/input_generators/input_generators.py#L44 )

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +159,43 @@ def __iter__(self):
...
+            self._repetitions = -1
+            self._len = repetitions
+
+    def __len__(self) -> int:
+        if self._len < 0:
+            raise LengthIsInfiniteException("The length is infinite!")
+        return self._len
+
+    def __iter__(self):
+        if self._repetitions == -1:
```
kralka, line 195: Bug with the following configuration
```
{       
        "operation": "repeat",
        "repetitions": -2,
        "configuration": {
            "operation": "constants",
            "name": "key",
            "values": [1, 2, 3]
        }
    }
```

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -145,3 +146,213 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+        "repetitions": -1,
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [1]
+        }
+    }
+    output = []
+    i = 0
+    for value in build_attack_points_iterator(config):
```
kralka, line 279: One can use `enumerate` instead of creating and updating a variable `i`.

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -145,3 +146,213 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+        output.append(value)
+        i += 1
+        if i > 5:
+            break
+
+    assert output == list(
+        iter(build_attack_points_iterator(config["configuration"]))) * i
```
kralka, line 286: This line fails when `config["configuration"]["values"]` is longer than one. And it is a bug in the test. One could use `zip` with `itertools.cycle` and compare elements in the for loop directly.

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -145,3 +146,213 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+    }
+    output = []
+    i = 0
+    for value in build_attack_points_iterator(config):
+        output.append(value)
+        i += 1
+        if i > 5:
+            break
+
+    assert output == []
```
kralka, line 307: This is good. What is unexpected is that `len(build_attack_points_iterator(config))` raises `LengthIsInfiniteException` even when nothing is iterated.

kralka (commented):
Thank you, please take a look at the comments

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -145,3 +147,247 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+
+
+def test_attack_point_iterator_repeat_infinite():
+    config = {
+        "operation": "repeat",
+        "repetitions": -1,
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [1]
```
kralka, line 275: Please either have a longer list here or create a similar test with more values.

kralka (commented):
PTAL

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -125,7 +132,8 @@ def get_generated_keys(self) -> List[str]:
 
 class AttackPointIteratorUnrestrictedGenerator(AttackPointIterator):
     """
-    Attack point iterator class that iterates over the unrestricted generator.
+    This exception is raised when the `__len__` function is called
```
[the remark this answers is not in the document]
LucaTomasko: should I make an extra file for it?

tests/capture/input_generators/test_attack_point_iterator.py
LucaTomasko: There is the test with repetitions: -2 that has values: [1, 2]

scaaml/capture/input_generators/attack_point_iterator.py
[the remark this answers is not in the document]
kralka: In general that is a good idea when a single file is getting larger and larger, but please don't do it in this PR.

Nevertheless the original comment was pointing out that you copy-pasted another comment into the wrong place (here instead of `LengthIsInfiniteException`).

tests/capture/input_generators/test_attack_point_iterator.py
kralka: Thank you for that other test. But this test as it is written fails when at line 275 you add more values to the list. And it is a bug in the test itself because on line 288 you multiply by the number of iterations `stop_count`. I am ok here with line 286 having a hardcoded expected value. But the current unit-test is rather confusing (as in when I change something it breaks and I would be blaiming the implementation not the test).

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +160,56 @@ def __iter__(self):
...
+        """Initialize the repeated iterate if repetitions is not present
+          or set to a negative number it will do an infinite loop 
+          if it is 0 it will not repeat at all.
+          
+          Args:
+            operation (str): The operation of the iterator
+                this gets asserted at the start.
+            configuration (Dict): The config for the iterated object
+                that will get repeated.
+            repetitions (int): This parameter decides how often the
```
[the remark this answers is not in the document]
kralka: Please substitute "how often" by "how many times".

https://www.gnu.org/software/sed/manual/sed.html

kralka (commented):
Please be sure to check the remaining comments. You can find a nice view at https://github.com/google/scaaml/pull/200/files

scaaml/capture/input_generators/attack_point_iterator.py
kralka: How was the original comment resolved? One does not "call" an exception, one "raises" an exception. At least that is the language used by https://docs.python.org/3/tutorial/errors.html Is there a reason why the docstring is saying that "exception get's called"?

kralka (commented):
Conversation resolved probably by mistake.

scaaml/capture/input_generators/attack_point_iterator.py
LucaTomasko: Sorry I changed it in the wrong place and thought is was resolved because of that

kralka (commented):
Thank you, LGTM from me but I have read it too many times, requesting another pair of eyes (JM).

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -152,3 +159,59 @@ def __iter__(self):
...
+    """
+    Attack point iterator class that iterates 
+    over a configuration a repeat amount of times.
+    """
+
+    def __init__(self,
+                 operation: str,
+                 configuration: Dict,
+                 repetitions: int = -1) -> None:
+        """Initialize the repeated iterate if repetitions is not present
```
jmichelp, line 174: Missing dot somewhere maybe? The sentence seems unusually long without punctuation

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -145,3 +147,245 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+        }
+    }
+    count = 0
```
jmichelp, line 278: Variable `count` is effectively unused in this test.
Consider rewriting as a for-loop
```python
for _ in range(5):
 for value1, value2 in zip(...):
  assert value1 == value2
```

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -145,3 +147,245 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+    config = {
+        "operation": "repeat",
+        "repetitions": -2,
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [1, 2]
+        }
+    }
+    count = 0
```
jmichelp, line 299: Same as previous comment

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -145,3 +147,245 @@ def test_attack_point_iterator_unrestricted_generator_all_args_len():
...
+        "configuration": {
+            "operation": "constants",
+            "name": "key",
+            "values": [[1], [2], [3]]
+        }
+    }
+
+    max_growth_factor = 2
+    python_process = psutil.Process()
+    memory_before_iterator = python_process.memory_info().rss
```
jmichelp, line 370: nit: consider pre-computing `memory_threshold = memory_before_iterator * max_growth_factor` instead of computing it for each assertion

tests/capture/input_generators/test_attack_point_iterator.py
jmichelp: Forgot it was an infinite loop here....

LucaTomasko: I don't think this works because the second for loop is an infinite loop

kralka (approved):
Thank you

PR #203, opened 2024-02-01
Bump action versions

jmichelp:
This removes the warnings about Node.js 16 being used.

Also put a scheduled run of the code coverage reports to ensure the badge always work.


kralka (approved):
Thank you!

PR #204, opened 2024-02-02
Code quality improvements.

jmichelp:


tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -1,3 +1,16 @@
+# Copyright 2024 Google LLC
```
kralka, line 1: QQ I have seen a lot of "Copyright 20CD-2024" Is that the preferred style, both are ok, or this one is preferred?

update_dependency_licenses.py
```diff
@@ -54,27 +70,53 @@ def GetPackageLicenses(package_name):
...
+    homepage: str
+
+    def __lt__(self, other: "PackageInfo") -> bool:
+        return self.licence < other.licence
+
+    def __iter__(self):
+        return iter((self.package, self.licence, self.category, self.homepage))
+
+    @classmethod
+    def get_headers(cls) -> Tuple[str, ...]:
```
kralka, line 100: Nit `Tuple[str, str, str, str]` tells me how many strings to expect.

update_dependency_licenses.py
```diff
@@ -54,27 +70,53 @@ def GetPackageLicenses(package_name):
...
+    category: str
+    homepage: str
+
+    def __lt__(self, other: "PackageInfo") -> bool:
+        return self.licence < other.licence
+
+    def __iter__(self):
+        return iter((self.package, self.licence, self.category, self.homepage))
+
+    @classmethod
```
kralka, line 99: Nit Any reason why this is not a `@staticmethod`? The argument `cls` seems to be unused and this would trigger pylint if we ever decide to lint outside the scaaml directory.

kralka (commented):
Thank you

tests/capture/input_generators/test_attack_point_iterator.py
jmichelp: When you first put the header, you put the current year. When you edit the file, you change the header to `20xx-20yy` with `yy` being the current year.
Does that answer your question?

update_dependency_licenses.py
jmichelp: I actually did that on purpose.
As the comment suggests, `get_headers()` and `__iter__()` must be kept in sync and the code will self adjust to how many things we return (compared to having headers hard-coded in the caller which requires a stricter type checking)

tests/capture/input_generators/test_attack_point_iterator.py
kralka: Totally, both is fine with me.

update_dependency_licenses.py
kralka: Ack

kralka (commented):
Ack the two comments

update_dependency_licenses.py
jmichelp: Fixed.

kralka (approved):
Awesome, thank you. Now orange is the new grey :)

Issue #85, opened 2022-07-06
mypy is questionable


kralka:
It seems that mypy has not discovered wrong return type typehint in [scaaml.io.Dataset.as_tfdataset](https://github.com/google/scaaml/blob/8b7e5ff700005c393f48aa1b22530f0f5a784bc4/scaaml/io/dataset.py#L380).

There are two issues:

1. Fixing the typehint.
2. Debugging and fixing mypy typechecking.

Python interpreter types the output as a tuple, when unpacking the types are tf.data.PrefetchDataset, dict, dict. The proper type should be Tuple[tf.data.Dataset, Dict, Dict]. Definitely not a Union of those.

jmichelp:
I guess that PR #207 should fix this. Although it's currently being blocked by fixing existing code base to comply with the new configuration.

kralka:
TODO (so that we do not forget). There are some files with:

# mypy: ignore-errors

https://github.com/google/scaaml/pull/207 will look good, but it will ignore those files (do not close this issue before we are happy with the ignored errors).

PR #209, opened 2024-02-08
First wave of fixes for the new mypy configuration file

jmichelp:
Required before merging #207 

Related to #85

### DIFF
--- a/scaaml/aes_forward.py
+++ b/scaaml/aes_forward.py
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 2022-2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Provides attack points for AES."""
+
+from collections.abc import Callable
+from typing import cast
 
 
 class AESSBOX:
@@ -96,7 +99,8 @@ class AESSBOX:
         return plaintext
 
     @classmethod
-    def get_attack_point(cls, name: str, **kwargs: bytearray) -> bytearray:
+    def get_attack_point(cls, name: str, key: bytearray,
+                         plaintext: bytearray) -> bytearray:
         """Return the correct attack point.
 
         Typical usage example:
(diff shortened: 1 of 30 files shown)


scaaml/aes_forward.py
```diff
@@ -109,4 +112,6 @@ def get_attack_point(cls, name: str, **kwargs: bytearray) -> bytearray:
         function = getattr(cls, name, None)
         if not function or not callable(function):
             raise ValueError(f"{name} is not a defined attack point")
-        return function(**kwargs)
+        function = cast(Callable[[bytearray, bytearray], bytearray], function)
```
kralka, line 115: NIT (by all means feel free to ignore / leave that for me or another PR) The parameter `name` has type `str` which is nice since as soon as we add another method we do not need to change this method. But it also means we are losing typechecking especially when there would be more methods. WDYT `assert type(function) is Callable[[bytearray, bytearray], bytearray]`?

jmichelp: We don't really loose anything. Type checking in Python is static and by default nothing happens at runtime.
The cast operation is only to make the typechecker happy.
But we can add runtime type checking too if you want.
I'd be more worried about the input generators though. I started typing them and it still uses Any because I wasn't sure about the types we want.

kralka (dismissed):
Thank you!

scaaml/aes_forward.py
kralka: Agree we are not losing anything, we could gain here a little bit, but I don't think that is necessary (very unlikely and the error message would be similar).

Ad input generators definitely planned to consolidate bytearrays and np arrays when moving to those.

scaaml/capture/scope/lecroy/lecroy_communication.py
```diff
@@ -173,27 +177,25 @@ def get_waveform(
...
         """Query binary data."""
         assert self._scope
         self._logger.debug("query_binary_values(message=\"%s\")", message)
-        return self._scope.query_binary_values(
+        raw_data = self._scope.query_binary_values(
             message,
             datatype=datatype,
             container=container,
         )
+        return cast(T, raw_data)
```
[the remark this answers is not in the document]
jmichelp: The PyVISA type is somehow broken. PyVISA type hints says that it returns `Sequence[Union[int, float]]` no matter the value of `container` (which is typed as `Union[Type, Callable[[Iterable], Sequence]]`).
And because `container` can be `Type`, it then can be anything, which will be called to build and return the results. That's what we're using to parse/build a `LeCroyWaveform`. 
So here I'm fixing the type hints by saying that `container` is a type and the function will return an instance of that type.
If we want to restrict the possible types, it can be done by specifying them at declaration of the `TypeVar`

kralka (approved):
Thank you

PR #211, opened 2024-02-14
More Mypy fixes

jmichelp:
Related to #85

### DIFF
--- a/scaaml/utils.py
+++ b/scaaml/utils.py
@@ -19,7 +19,8 @@ from typing import Any, Dict, List, Sequence
 import time
 
 from glob import glob
-from termcolor import _types, cprint
+from termcolor import cprint
+from termcolor._types import Color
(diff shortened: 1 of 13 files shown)


scaaml/utils.py
```diff
@@ -15,17 +15,19 @@
 
 from multiprocessing import Pool
 from random import randint
+from typing import Any, Dict, List, Sequence
 import time
 
 from glob import glob
-from termcolor import cprint, termcolor
+from termcolor import _types, cprint
```
kralka, line 22: NIT for readability I would prefer (since `_types` is rather generic).
```
from termcolor._types import Color
from termcolor import cprint
```

kralka (approved):
Thank you, LGTM (one nit, feel free to ignore)

kralka (approved):
Thank you

PR #212, opened 2024-02-14
Implement zip iterator

LucaTomasko:


tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -457,3 +457,256 @@ def test_repeat_memory():
...
+    output = build_attack_points_iterator(config)
+
+    assert output.get_generated_keys() == [
+        config["operands"][0]["name"], config["operands"][1]["name"],
+        config["operands"][2]["name"]
+    ]
+
+
+def test_attack_point_iterator_zip_no_operands():
+    values = [[0], [1], [2]]
```
kralka, line 705: NIT Unused

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -238,3 +239,48 @@ def __iter__(self) -> AttackPointIteratorT:
...
+            except LengthIsInfiniteException:
+                self._len = -1
+
+    def __len__(self) -> int:
+        if self._len < 0:
+            raise LengthIsInfiniteException("The length is infinite!")
+        return self._len
+
+    def __iter__(self) -> AttackPointIteratorT:
+        return iter(
```
kralka, line 276: Honestly I had quite hard time reading this (three for loops in an iterator expression is too many for me). 
WDYT about:

```
# zip will return a tuple of output dictionaries, we flatten them.
return iter(merge_dictionaries(tuple_of_dictionaries) for tuple_of_dictionaries in zip(*self._operands))
```

With a function `merge_dictionaries` which gets a tuple of dictionaries and uses `update` method of a Python dictionary to merge them (similar to `get_generated_keys` below). This way we can easily unit-test the `merge_dictionaries`.

kralka (commented):
Thank you!

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -238,3 +241,53 @@ def __iter__(self) -> AttackPointIteratorT:
...
+                has to be in the config file. This is only used once to
+                double check if the operation is the correct one.
+                
+            operands (List[Dict[str, Any]]): The operands are any number of
+                iterator configs that will be combined."""
+        assert "zip" == operation
+        self._operands = list(
+            build_attack_points_iterator(operand) for operand in operands)
+
+        list_of_lengths = [
```
kralka, line 265: NIT
```suggestion
        non_negative_lengths = [
```

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -238,3 +241,53 @@ def __iter__(self) -> AttackPointIteratorT:
...
+            operand._len for operand in self._operands if operand._len >= 0
+        ]
+        if not self._operands:
+            self._len = 0
+        else:
+            self._len = min(list_of_lengths, default=-1)
```
kralka, line 271: NIT
```suggestion
            # If `non_negative_lengths` is empty it means that all operands are infinite.
            self._len = min(non_negative_lengths, default=-1)
```

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -238,3 +241,53 @@ def __iter__(self) -> AttackPointIteratorT:
...
+
+    def __len__(self) -> int:
```
kralka, line 273: NIT This implementation is repeated from the repeat iterator. It would also work for all others since we cache `self._len` and return it (there is an additional check for being non negative, but that is negligible). I would implement `__len__` only once on the base method.

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -238,3 +241,53 @@ def __iter__(self) -> AttackPointIteratorT:
...
+        if self._len < 0:
+            raise LengthIsInfiniteException("The length is infinite!")
+        return self._len
+
+    def __iter__(self) -> AttackPointIteratorT:
+        return iter(
+            self.merge_dictionaries(tuple_of_dictionaries)
+            for tuple_of_dictionaries in zip(*self._operands))
+
+    def merge_dictionaries(self, tuple_of_dictionaries) -> Dict:
```
kralka, line 283: NIT
```suggestion
    @staticmethod
    def merge_dictionaries(tuple_of_dictionaries: Tuple[Dict[str, List[int]]]) -> Dict[str, List[int]]:
```

tests/capture/input_generators/test_attack_point_iterator.py
```diff
@@ -457,3 +457,329 @@ def test_repeat_memory():
...
+    for value1, value2 in zip(
+            build_attack_points_iterator(config),
+            zip(itertools.cycle(values), itertools.cycle(values))):
+        if count > 4:
+            break
+        count += 1
+        assert value1 == {
+            config["operands"][0]["configuration"]["name"]: value2[0],
+            config["operands"][1]["configuration"]["name"]: value2[1]
+        }
```
kralka, line 685: NIT please also add `with pytest.raises(LengthIsInfiniteException)` for `len`.

kralka (commented):
Thank you, looks very nice now. Just a couple nits to make it even better :)

PR #215, opened 2024-02-15
More mypy stuff...

jmichelp:
Continuation of fixing Issue #85 

After this PR, all `scaaml/` is done except `scaaml/intro`

### DIFF
--- a/scaaml/io/dataset.py
+++ b/scaaml/io/dataset.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Google LLC
+# Copyright 2021-2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -13,9 +13,6 @@
 # limitations under the License.
 """Build and load tensorFlow dataset Record wrapper"""
 
-# Temporarily ignore this file until it's properly typed or reworked
-# mypy: ignore-errors
-
 import copy
 import math
 import json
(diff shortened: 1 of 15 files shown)


scaaml/io/dataset.py
```diff
@@ -40,11 +39,28 @@
...
 import scaaml.io.utils as siutils
 from scaaml.io.utils import dtype_name_to_dtype, dtype_dtype_to_name
-from scaaml.io.shard import Shard
+from scaaml.io.shard import Shard, CompressionT
 from scaaml.io.errors import DatasetExistsError
 
+_T = TypeVar("_T")
+
+
+class FakeTqdm(Generic[_T]):
```
kralka, line 48: NIT This is not necessary, tqdm has a parameter `disable: bool`

scaaml/io/dataset.py
```diff
@@ -362,26 +387,23 @@ def close_shard(self):
...
-                     batch_size: int = 32,
-                     prefetch: int = 4,
-                     file_parallelism: Optional[int] = os.cpu_count(),
-                     parallelism: Optional[int] = os.cpu_count(),
-                     shuffle: int = 1000,
-                     additional_attack_points: Optional[List[Dict]] = None,
-                     **kwargs) -> Union[tf.data.Dataset, Dict, Dict]:
+    def as_tfdataset(
+        dataset_path: str,
+        split: str,
```
kralka, line 392:
```suggestion
        split: SPLIT_T,
```

kralka (approved):
Thank you

scaaml/io/dataset.py
jmichelp: That would require more code changes though because the function that builds the progressbar takes a class instead of a bool as a parameter.
I prefer leaving this to another PR, this one only deals with type checking.

scaaml/io/dataset.py
jmichelp: Done

kralka (approved):

PR #218, opened 2024-02-22
Implement cartesian product iterator

LucaTomasko:
Issue #196


scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +280,57 @@ def get_generated_keys(self) -> List[str]:
...
+        assert "cartesian_product" == operation
+        self._operands = list(
+            build_attack_points_iterator(operand) for operand in operands)
+
+        operand_lengths = [operand._len for operand in self._operands]
+        if not self._operands or any(
+                length <= 0
+                for length in operand_lengths) or len(operand_lengths) != 2:
+            self._len = 0
+            print(len(self))
```
kralka, line 311: Remove debugging print.

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +280,57 @@ def get_generated_keys(self) -> List[str]:
...
+            operands (List[Dict[str, Any]]): The operands are any number of
+                iterator configs that will be combined."""
+        assert "cartesian_product" == operation
+        self._operands = list(
+            build_attack_points_iterator(operand) for operand in operands)
+
+        operand_lengths = [operand._len for operand in self._operands]
+        if not self._operands or any(
+                length <= 0
+                for length in operand_lengths) or len(operand_lengths) != 2:
```
kralka, line 309: A cartesian product of a single operand should iterate that operand.

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +280,57 @@ def get_generated_keys(self) -> List[str]:
...
+            self._len = 0
+            print(len(self))
+        else:
+            self._len = operand_lengths[0] * operand_lengths[1]
+
+    def __iter__(self) -> AttackPointIteratorT:
+        if any(len(operand) <= 0 for operand in self._operands) or len(
```
kralka, line 316: This line raises if any of the operands is infinite. This fact is not in a docstring.

The check for infinite iterations should happen in __init__.

Technically we could decide to be ok with possibly the first operand having infinite length (up to you, since this would make the code more complex).

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +280,57 @@ def get_generated_keys(self) -> List[str]:
...
+                self._operands) != 2:
+            raise ValueError
+        return iter(
+            self.merge_dictionaries(value_one, value_two)
+            for value_one in self._operands[0]
+            for value_two in self._operands[1])
+
+    @staticmethod
+    def merge_dictionaries(value_one: Dict[str, Any],
```
kralka, line 325: Why is this method being copy-pasted? It can easily go to the parent class. Also please rename it to `_merge_dictionaries` since the user is not expected to call it.

scaaml/capture/input_generators/attack_point_iterator.py
kralka: If there are three or more operands you can do an easy trick:

```
self._operands = [build_attack_points_iterator(operands[0]), build_attack_points_iterator({"operation": "cartesian_product", "operands": operands[1:]})]
```

This requires some refactoring here and there. Also I believe that unit-tests should have three or four operands and they currently don't have those.

kralka (commented):
Thank you, please take a look at the comments

scaaml/capture/input_generators/attack_point_iterator.py
LucaTomasko: But it isn't used by every child class is it ok to put it in the parent then?

scaaml/capture/input_generators/attack_point_iterator.py
LucaTomasko: what is the difference between this to the original one?

scaaml/capture/input_generators/attack_point_iterator.py
kralka: Fine by me, definitelly less wrong than duplicating code. You can also make it a free function if you wish.

scaaml/capture/input_generators/attack_point_iterator.py
LucaTomasko: Oh ok I see but then you would have to check if length is longer then 1 right?

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +281,66 @@ def get_generated_keys(self) -> List[str]:
...
+        elif not self._operands or any(
+                length == 0 for length in operand_lengths):
+            self._len = 0
+        else:
+            self._len = math.prod(operand_lengths)
+
+    def __iter__(self) -> AttackPointIteratorT:
+        if len(self._operands) == 2:
+            return iter(
+                self._merge_dictionaries(value_one, value_two)
```
kralka, line 328: The `_merge_dictionaries` is pretty much a copy-paste.
```suggestion
                {**value_one, **value_two}
```

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +281,68 @@ def get_generated_keys(self) -> List[str]:
...
+        """Initialize the cartesian product iterator.
+          
+          Args:
+            operation (str): The operation of the iterator
+                represents what the iterator does and what 
+                has to be in the config file. This is only used once to
+                double check if the operation is the correct one.
+                
+            operands (List[Dict[str, Any]]): The operands are any number of
+                iterator configs that will be combined."""
```
kralka, line 302: Raises part of the docstring is missing.

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +281,70 @@ def get_generated_keys(self) -> List[str]:
...
+    def __init__(self, operation: str, operands: List[Dict[str, Any]]) -> None:
+        """Initialize the cartesian product iterator.
+          
+          Args:
+            operation (str): The operation of the iterator
+                represents what the iterator does and what 
+                has to be in the config file. This is only used once to
+                double check if the operation is the correct one.
+                
+            operands (List[Dict[str, Any]]): The operands are any number of
```
kralka, line 301: - Zero is a number, but it still raises. This is not documented.
- What type of exception/error is raised?

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +281,70 @@ def get_generated_keys(self) -> List[str]:
...
+                iterator configs that will be combined. If the iterator
```
kralka, line 302: The documentation is a little vague. It raises whenever any of the operands would result in an infinite iterator. Not the same as when the whole is infinite (product of infinite and zero length would be zero length but this still raises).

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +281,75 @@ def get_generated_keys(self) -> List[str]:
...
+                **value_one,
+                **value_two
+            }
+                        for value_one in self._operands[0]
+                        for value_two in self._operands[1])
+        else:
+            return iter(self._operands[0])
+
+    @staticmethod
+    def _merge_dictionaries(value_one: Dict[str, Any],
```
kralka, line 344: Where is this method used?

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +281,67 @@ def get_generated_keys(self) -> List[str]:
...
+            operation (str): The operation of the iterator
+                represents what the iterator does and what 
+                has to be in the config file. This is only used once to
+                double check if the operation is the correct one.
+                
+            operands (List[Dict[str, Any]]): The operands are any number of
+                iterator configs that will be combined. If the operands list
+                is empty it will raise a ValueError. If one of the operands
+                length is 0 the length of the cartesian product iterator will
+                also be 0, it will return an empty iterator.If one of the
```
kralka, line 305: Missing space in front of the "iterator.If"

scaaml/capture/input_generators/attack_point_iterator.py
```diff
@@ -279,3 +281,67 @@ def get_generated_keys(self) -> List[str]:
...
+                build_attack_points_iterator(operands[0]),
+                build_attack_points_iterator({
+                    "operation": "cartesian_product",
+                    "operands": operands[1:]
+                })
+            ]
+        else:
+            self._operands = [build_attack_points_iterator(operands[0])]
+        operand_lengths = [operand._len for operand in self._operands]
+        if not self._operands or any(length == 0 for length in operand_lengths):
```
kralka, line 323: The `not self._operands` is not used.

PR #221, opened 2024-02-28
Move from coveralls-python to Github action

jmichelp:
Last release of coveralls-python is from 2021.
Removing that dependency before things starts to break.

The Github action doesn't handle the python `.coverage` file format very well so we convert it to XML before.

I considered adding a problem matcher too but that requires a lot of work as:
  - Github limits the amount of problems that can reported per test
  - coverage output isn't really in a machine-friendly format
  - problems only apply to a single line, which wouldn't help correctly reporting the untested line
  - we would need to filter out files that are not part of the PR

Conclusion is: let's use coveralls web UI instead :)

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -61,13 +61,15 @@ jobs:
         run:
           pip install --upgrade picosdk
       - name: Install workflow dependencies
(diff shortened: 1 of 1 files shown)


kralka (approved):
Thank you

PR #223, opened 2024-03-06
Stop supporting Python3.8

wsxrdv:


scaaml_intro/train.py
```diff
@@ -31,8 +31,9 @@
 
 def train_model(config):
     tf_cap_memory()
-    train_glob = f"datasets/{config['algorithm']}/train/*"
-    test_glob = f"datasets/{config['algorithm']}/test/*"
+    algorithm = config['algorithm']
```
jmichelp: surprised pylint doesn't complain about not using double quotes here.

kralka: It was, but not anymore. PTAL

jmichelp (approved):

PR #226, opened 2024-03-20
Implement constant and generated pydantic model

LucaTomasko:


scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,117 @@
...
+
+from scaaml.capture.input_generators.attack_point_iterator_exceptions import ListNotPrescribedLengthException
+from scaaml.capture.input_generators.input_generators import balanced_generator, unrestricted_generator
+
+AttackPointIteratorT = Iterator[Dict[str, Any]]
+
+
+class ConstantIteratorModel(BaseModel):
+    """
+    Attack point iterator pydantic model that iterates over a constant.
```
kralka, line 26:
```suggestion
    Attack point iterator pydantic model that iterates over given constants.
```

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,117 @@
...
+    
+    Args:
+        operation (str): The operation of the iterator
+        represents what the iterator does and what 
+        arguments should be present. This is only used once to
+        double check if the operation is the correct one.
```
kralka, line 32: The last sentence does not make much sense now. Since it is also used by the tagged union.

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,117 @@
...
+        name (str): The name represents the key name of the value.
+
+        length (int): The prescribed length for each list in values.
+        If one of the lists isn't the same length as this variable
+        it will raise an ListNotPrescribedLengthException.
+
+        values (List[List[int]]): List of lists of ints that gets
+        iterated through.
+    """
+    operation: Literal['constants']
```
kralka, line 43: Move down and put a default value.
```suggestion
    operation: Literal["constants"] = "constants"
```

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,117 @@
...
+    length (int): Length of each yielded array for the generator.
+
+    bunches (int): How many bunches to return (yields `bunches * elements`
+        random arrays) for the generator.
+    
+    elements (int): Each array contains numbers in `range(elements)`
+        for the generator.
+    """
+
+    operation: Literal['balanced_generator', 'unrestricted_generator']
```
kralka, line 89: NIT throughout this file. Why are we switching to using ' instead of "?
```suggestion
    operation: Literal["balanced_generator", "unrestricted_generator"]
```

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,117 @@
...
+    @model_validator(mode='after')
+    def check(self) -> 'ConstantIteratorModel':
+        for value in self.values:
+            if len(value) != self.length:
+                raise ListNotPrescribedLengthException(
+                    f'The prescribed length is {self.length} and \
+                    the length of {value} is {len(value)}.')
+        return self
+
+    def len(self) -> int:
```
kralka, line 57: What is the reason to implement `len` and not `__len__`?

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,117 @@
...
+        """Return the number of iterated elements."""
+        return len(self.values)
+
+    def iter(self) -> AttackPointIteratorT:
```
kralka, line 61: What is the reason to implement `iter` and not `__iter__`?

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,117 @@
...
+        for the generator.
+    """
+
+    operation: Literal['balanced_generator', 'unrestricted_generator']
+    name: str
+    length: int
+    bunches: int = 1
+    elements: int = 256
+
+    def len(self) -> int:
```
kralka, line 95: Same here.

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,117 @@
...
+        """Return the number of iterated elements."""
+        return self.bunches * self.elements
+
+    def iter(self) -> AttackPointIteratorT:
```
kralka, line 99: Same here.

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,117 @@
...
+        if self.operation == 'balanced_generator':
+            return iter({self.name: value}
```
kralka, line 101: Functions are first-class objects in Python. You can do:

```
if self.operation == "balanced_generator":
  generator = balanced_generator
elif self.operation == "unrestricted_generator":
  generator = unrestricted_generator
else:
  raise ValueError(f"Unknown generator type: {self.operation}")

return iter({self.name: value} for value in generator(length=....
```

kralka (commented):
Thank you.

General question: is this code ever used? As in are there at least some unit-tests to see it works? If not and this is in fact a dead code until everything is implemented we should probably consider creating a separate branch with a couple of commits and then squashing those before merging to main?

scaaml/capture/input_generators/attack_point_iterator_models.py
LucaTomasko: If I use __iter__ it conflicts with the BaseModel __iter__ and says that the return type is not compatible

scaaml/capture/input_generators/attack_point_iterator_models.py
LucaTomasko: cause I wanted to make it consistent with iter and some future len's will have negative return values and that is not compatible with __len__. The used model will have __len__ though

kralka: I thought that `len` which calls `__len__` does not like negative numbers and too large numbers.

```
>>> class A:
...   def __len__(self):
...     return -1
... 
>>> a = A()
>>> len(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: __len__() should return >= 0
>>> a.__len__()
-1
```

But having `my_obj.len()` without having `my_obj.__len__()` or `len(my_obj)` sounds counterintuitive. I would prefer to keep it as close to https://github.com/google/scaaml/blob/main/scaaml/capture/input_generators/attack_point_iterator.py as possible.

NIT rather hard to read when you don't use the backticks. `__len__` vs __len__ :)

scaaml/capture/input_generators/attack_point_iterator_models.py
kralka: Ack, not an easy decision here. The problem I see is that both of the following are allowed:

```
const = ConstantIteratorModel(operation"constants", name="plaintext", length=3, values=[[1, 2, 3]])

for t in const:  # I will write this in the future.
  print(t)

for t in const.iter():
  print(t)
```

I am not sure what happens when we overwrite the `BaseModel.__iter__` since it might be used internally, e.g., in serialization.

kralka: @jmichelp solution: Let us call that function `items` to keep it closer to Python dictionaries. But please add documentation string so that people use that.
```suggestion
    def items(self) -> AttackPointIteratorT:
```

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,115 @@
...
+        values (List[List[int]]): List of lists of ints that gets
+        iterated through.
+    """
+    operation: Literal["constants"] = "constants"
+    name: str
+    length: int
+    values: List[List[int]]
+
+    @model_validator(mode="after")
+    def check(self) -> "ConstantIteratorModel":
```
jmichelp, line 48: I guess `Self` wasn't working here as a return type?

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class GeneratedIteratorModel(BaseModel):
+    """
+    Attack point iterator pydantic model that iterates over
+    the balanced and unrestricted generator.
+
+    operation (str): The operation of the iterator
+        represents what the iterator does and what 
+        arguments should be present.
```
jmichelp, line 74: nit: Because this class accepts multiple values for `operation`, I think adding the possible values and what they mean in the docstring makes sense.

scaaml/capture/input_generators/attack_point_iterator_models.py
LucaTomasko: Nope

jmichelp (approved):

PR #227, opened 2024-03-21
Fix mypy missing type parameters for keras.Model

kralka:


kralka:
@jmichelp PTAL this was blocking dependabot (and likely other PRs).

jmichelp (approved):

PR #234, opened 2024-04-04
Add zip, cartesian product and general iterator pydantic model

LucaTomasko:


scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -165,3 +166,177 @@ def items(self) -> AttackPointIteratorT:
...
+            RepeatIteratorModels that will be combined.
+    """
+    operation: Literal["zip"]
+    operands: List[Union[BasicIteratorModels, RepeatIteratorModel]]
+
+    def __len__(self) -> int:
+        non_negative_lengths: List[int] = []
+        for operand in self.operands:
+            try:
+                if len(operand) >= 0:
```
kralka, line 193: This always holds.

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -165,3 +166,177 @@ def items(self) -> AttackPointIteratorT:
     def get_generated_keys(self) -> List[str]:
         """Returns an exhaustive list of names this iterator will create."""
         return self.configuration.get_generated_keys()
+
+
+class ZipIteratorModel(BaseModel):
+    """
+    Attack point iterator zip pydantic model. This class takes any amount of
+    operands and combines them just like the `zip` function in Python.
```
kralka, line 174: This docstring is a little misleading. Because `zip` iterates over tuples. However this iterates over dictionaries.

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -165,3 +166,177 @@ def items(self) -> AttackPointIteratorT:
...
+    @model_validator(mode="after")
+    def check_model(self) -> "CartesianProductIteratorModel":
+        if len(self.operands) > 1:
+            self.operands = [
+                self.operands[0],
+                CartesianProductIteratorModel(operation="cartesian_product",
+                                              operands=self.operands[1:])
+            ]
+        else:
+            self.operands = [self.operands[0]]
```
kralka, line 265: I do not really understand this line. Could we have a documentation for it? I am not even sure it does what it is supposed to do.

kralka (commented):
Thank you

scaaml/capture/input_generators/attack_point_iterator_models.py
LucaTomasko: it does what it is supposed to do. We have the same code in the original attack point iterators and I tested this one to see if it works.

LucaTomasko: I guess this line doesn't really do anything

kralka: Well it does. It raises an exception when `self.operands` is an empty list. Which is not very user-friendly (kind of unexpected exception somewhere deep in the code, no descriptive message).

scaaml/capture/input_generators/attack_point_iterator_models.py
kralka: `len` always returns a non-negative number. That is why there is the `InfiniteLengthException`. So this if condition is not needed (just the line inside it would be doing exactly the same).

scaaml/capture/input_generators/attack_point_iterator_models.py
LucaTomasko: Does it because I thought pydantic makes you input something or does it allow empty lists []?

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -165,3 +166,179 @@ def items(self) -> AttackPointIteratorT:
...
+
+    def __len__(self) -> int:
+        operand_lengths: List[int] = []
+        for operand in self.operands:
+            try:
+                operand_lengths.append(len(operand))
+            except LengthIsInfiniteException:
+                operand_lengths.append(-1)
+
+        if any(length == 0 for length in operand_lengths):
```
kralka, line 279: NIT
```suggestion
        if 0 in operand_lengths:
```

kralka (approved):
Thank you, LGTM. If you want to fix the two NIT feel free. Otherwise I am submitting.

kralka (approved):
Thank you

PR #236, opened 2024-04-11
Update tests to new model

LucaTomasko:


kralka (commented):
IMO it would be a nice idea to move the `build_attack_points_iterator` to use the new `IteratorModel` and see that all old tests pass. This way we are only deleting code without having a replacement for it.

scaaml/capture/input_generators/attack_point_iterator_models.py
```diff
@@ -341,3 +342,9 @@ def __len__(self) -> int:
 
     def items(self) -> AttackPointIteratorT:
         return self.iterator_model.items()
+
+
+def build_attack_points_iterator(
+        configuration: Dict[str, Any]) -> IteratorModel:
+    configuration = copy.deepcopy(configuration)
```
kralka, line 349: NIT Probably we do not need to copy the configuration. Since Pydantic is unlikely to change it. Please check. Leaving it is also ok.
```suggestion
```

kralka (approved):
Thank you! Looks great, glad that unit-tests are passing. Left a tiny NIT, otherwise LGTM.

scaaml/capture/input_generators/attack_point_iterator_models.py
LucaTomasko: Does not look like it

kralka (approved):
Thank you!

PR #238, opened 2024-04-12
Add references to publications and datasets

kralka:
Please do not submit yet. Uploading datasets and models (and code snippets for loading them) is work in progress.


.github/workflows/mdlint.yml
```diff
@@ -0,0 +1,20 @@
...
+      - '**/*.md'
+      - '.markdownlint.json'
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  mdlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
jmichelp, line 14: `checkout@v4` to avoid NodeJS deprecation issue

kralka: ipynb and py files :/

jmichelp: According to the doc, you can put a `.markdownlintignore` file at the root of the repo to list the things to ignore (similar syntax at the `.gitignore` one). As that's processed by the underlying tool and not the action itself, it may work more reliably.

kralka: For some reason the `.markdownlintignore` refuses to ignore `README.md` (at least it ignores ipynb and py). But these two lines are a mistake:

.github/workflows/mdlint.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  mdlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: markdownlint-cli
+        uses: nosborn/github-action-markdown-cli@v1.1.1
+        with:
+          files: '**/*.md'
+          ignore_path: ".markdownlintignore"
```
jmichelp, line 19: This is not supported by v1.1.1
Deleting the line is fine as we're using the default filename anyway.
You may also bump the action to `@v3` tag instead of `v1.1.1`, as you did before.

.github/workflows/mdlint.yml
jmichelp: They are 253 characters long though and could be shortened but would reduce the readability of the table.
Probably some tweaks on the configuration file. This can wait a following PR as we're also missing the custom matcher to annotate the PR.

.github/workflows/mdlint.yml
kralka: Now papers/2024/GPAM/README.md:59 is not being linted :/

PR #266, opened 2024-08-13
Fix trigger buffer

kralka:
This should not happen. Not sure why it did. Hot-fix for now, added an issue to deal with this later.


kralka:
This is just a hotfix of https://github.com/google/scaaml/issues/267

PR #272, opened 2024-08-30
Made changes to some errors prone lines

shyam314159:
1.  Added code in `train.py` so one can load the model if it already exists and continue training.
2.  Added `.keras`  file extension so it shows no error while saving the model or loading the model.


scaaml_intro/train.py
```diff
@@ -77,17 +79,23 @@ def train_model(config):
...
+                # model recording setup
                 cb = [
                     ModelCheckpoint(monitor="val_loss",
-                                    filepath=f"models/{stub}",
+                                    filepath=file_path,
                                     save_best_only=True),
                     TensorBoard(log_dir="logs/" + stub, update_freq="batch")
                 ]
-
+                if os.path.exists(file_path):
```
kralka, line 92: Is this check needed? It seems more intuitive to "start training again" than to "train a certain number of epochs more".

scaaml_intro/train.py
```diff
@@ -77,17 +79,23 @@ def train_model(config):
             # multi gpu
             strategy = tf.distribute.MirroredStrategy()
             with strategy.scope():
-                model = get_model(input_shape, attack_point, config)
-
-                # model recording setup
                 stub = get_model_stub(attack_point, attack_byte, config)
+                file_path = f"models/{stub}.keras"
```
kralka, line 83: This modernises the code. However we would be silently losing backwards compatibility with pretrained models. Since this code is for the 2019 Defcon presentation I would prefer to pin a TensorFlow version here (say 2.12) and have more up to date code for the more recent projects. What do you think?

kralka (commented):
Thank you for your contribution.

scaaml_intro/train.py
shyam314159: It's indeed used for "start training again" as an individual I have no computational resources to continue 75 epochs in single go this will be helpful.

scaaml_intro/train.py
shyam314159: you're right back compatibility is a concern but when I tried to run the code it threw an error asking to add a keras extension so I did. I was using the latest version of everything.

shyam314159: ok this might not be an issue if we install the version mentioned in requirements.

scaaml_intro/train.py
shyam314159: yes I was talking about ASCAD dataset

scaaml_intro/train.py
shyam314159: it wasn't loading with no extension back ward compatibility is a concern but

ZetaMaster777 (approved):

kralka:
@shyam314159 do you plan to follow up on this PR?

- I have a todo for the SCA tutorial with GPAM and https://github.com/google/sedpack/pull/25
- I would vote for keeping backwards compatibility and keeping this code as it was (possibly with a note about the older version of TensorFlow).

shyam314159:
> @shyam314159 do you plan to follow up on this PR?
> 
> * I have a todo for the SCA tutorial with GPAM and [Add a side channel attack tutorial sedpack#25](https://github.com/google/sedpack/pull/25)
> * I would vote for keeping backwards compatibility and keeping this code as it was (possibly with a note about the older version of TensorFlow).

yes I'd love to, I am working to AI for side channel attack this would help me a lot, and I see the code for GPAM is also added that's quite interesting but currently am busy with some other work I'll get back to it as soon as I can.

kralka:
I would vote for closing this PR without merging.

- One can import the GPAM model (updated version) as `from scaaml.models import get_gpam_model` without breaking compatibility with old saved files. This also allows to save models during training and works with recent TF and Keras versions.
- There is a tutorial of using GPAM (with a naive set of hyperparameters -- hypertuning is left as an exercise) on the historical TinyAES dataset: https://google.github.io/sedpack/tutorials/sca/gpam/
- There is a note on which publication is which (the introduction demo versus GPAM): https://google.github.io/scaaml/papers/scaaml_defcon_2019/

@shyam314159 what do you think?

PR #285, opened 2024-09-16
Wrong trigger...

jmichelp:
As the comment was saying, it was to manually run and for the merge queue to work we need `merge_group` instead


kralka (approved):
Thank you!

PR #296, opened 2024-09-25
Implement SNR

wsxrdv:


scaaml/stats/online.py
```diff
@@ -0,0 +1,264 @@
...
+# limitations under the License.
+"""Online computation of statistical values."""
+
+from typing import Optional
+
+import numpy as np
+
+
+class Sum:
+    """Compute a sum in an online fashion in online fashion, numerically
```
jmichelp, line 22: redundant?

scaaml/stats/online.py
```diff
@@ -0,0 +1,264 @@
...
+    def __init__(self, ddof: int = 0) -> None:
+        """Initialize a sum.
+
+        Args:
+
+          ddof (int): Degrees of freedom. Defaults to 0. For Bessel's
+          correlation set to 1. For more information see `np.var`.
+        """
+        self._ddof: int = ddof
+        self._mean: Sum = Sum()
```
jmichelp, line 140: `mean` is `Sum`? That's counter-intuitive considering that in the 2-pass version, `mean` is `Mean`

scaaml/stats/snr.py
```diff
@@ -0,0 +1,156 @@
...
+          use_hamming_weight (bool): Use just the Hamming weight of the value.
+        """
+        assert byte_index in range(16)
+        self._hamming_weight_table: list[int] = self._precompute_hw_table()
+        self._byte_index: int = byte_index
+        self._use_hamming_weight: bool = use_hamming_weight
+        self._attack_point: AttackPoint = attack_point
+
+    @staticmethod
+    def _precompute_hw_table() -> list[int]:
```
jmichelp, line 57: The whole function seems to be a convoluted way to compute the same as:
```python
return [x.bit_count() for x in range(256)]
```

scaaml/stats/snr.py
```diff
@@ -0,0 +1,156 @@
...
+
+        return byte_value
+
+
+class SNRSinglePass:
+    """Single pass SNR implementation.
+    """
+
+    def __init__(self,
+                 leakage_model: LeakageModelAES128,
```
jmichelp, line 103: Maybe abstract AES away?
```python
class LeakageModel(Protocol):
  def leakage(self, **params: Any) -> int:
    ...

```

And use that as a type instead of `LeakageModelAES128`.
This way the leakage models might be defined elsewhere than in the SNR as they might be useful for other computations

scaaml/stats/snr.py
```diff
@@ -0,0 +1,156 @@
...
+            hwi = 0
+            x = i
+            while x > 0:
+                if x % 2 == 1:
+                    hwi += 1
+                x = x // 2
+            r.append(hwi)
+        return r
+
+    def leakage(self, plaintext: np.typing.NDArray[np.float64],
```
jmichelp, line 71: how `plaintext` and `key` be floats?

scaaml/stats/snr.py
```diff
@@ -0,0 +1,156 @@
...
+                                      f"{self._attack_point}")
+
+        # Maybe convert to Hamming weight.
+        if self._use_hamming_weight:
+            return self._hamming_weight_table[byte_value]
+
+        return byte_value
+
+
+class SNRSinglePass:
```
jmichelp, line 98: Should these classes inherit from a Protocol class to enforce the skeleton and be used in typing?

scaaml/stats/online.py
kralka: Yes it is a mean which we compute online. First update we set it to the value. Then we update it by (value - mean) / n_seen. Meaning this variable is a running mean without having to represent large numbers. I can update it to Mean, but then I need the delta anyway.

scaaml/stats/snr.py
wsxrdv: When we have another leakage (e.g., ECC) definitely. Now it felt like over-engineering. Happy to add it if you think.

scaaml/stats/.snr.py.swp
jmichelp, line 1: gitignore the `.swp` files please. Or make sure `vi` is closed before committing :)

scaaml/stats/snr.py
jmichelp: Ack

scaaml/stats/snr.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def update(
+        self, example: dict[str, Union[np.typing.NDArray[np.uint8],
+                                       np.typing.NDArray[np.float64]]]
+    ) -> None:
+        """Update itself with another example.
+
+        Args:
+
+          example (dict[str, np.typing.NDArray[np.float64]]): Assumes that
```
jmichelp, line 112: docstring not in sync with actual function declaration

jmichelp (approved):

kralka (approved):
Just added docstring, no need to bother JM again

PR #354, opened 2025-03-21
Errata for the GPAM model

wsxrdv:
Changes:

- ScaledNorm instead of BatchNorm. This is according to the paper (Figure 1 in merging).
- More dense layers in the head construction. This has been a human error during publishing the code.

We are grateful to the team (currently getting permission and a full list of names) for letting us know that they could not reproduce the results.


kralka:
The save & load test needs to be much faster. But still after at least some learning (x_train=x_train[:10_000], ..., single epoch).

kralka (approved):

PR #371, opened 2025-04-16
Point to ECC firmware code

kralka:
An equivalent to the ECC firmware used for the GPAM paper has been merged to the ChipWhisperer repository.

### DIFF
--- a/papers/datasets/ECC/GPAM/README.md
+++ b/papers/datasets/ECC/GPAM/README.md
@@ -76,3 +76,13 @@ ChipWhisperer repository
 ([commit
 6f0b3e0b0bc3f8423c6ee0428ae9712ad5412454](https://github.com/newaetech/chipwhisperer/commit/6f0b3e0b0bc3f8423c6ee0428ae9712ad5412454)).
 Note that we have not tested 1:1 compatibility with the pre-trained models.
+Notable differences:
+
+-   Use of the ChipWhisperer firmware stack vs bare-metal dedicated firmware
+    used in the paper. Impact should be negligible when the trigger is high but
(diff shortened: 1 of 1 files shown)


papers/datasets/ECC/GPAM/README.md
```diff
@@ -67,3 +67,12 @@ Multiplicative blinding using a 128-bit long random mask.
...
 protected by an independent CM2).
+
+## ECC Firmware
+
+A firmware equivalent to the one used for the GPAM paper has been merged to the
+ChipWhisperer repository
+[simpleserial-ecc.c](https://github.com/newaetech/chipwhisperer/blob/develop/firmware/mcu/simpleserial-ecc-notrace/simpleserial-ecc.c)
+([commit
+6f0b3e0b0bc3f8423c6ee0428ae9712ad5412454](https://github.com/newaetech/chipwhisperer/commit/6f0b3e0b0bc3f8423c6ee0428ae9712ad5412454)).
+Note that we have not tested 1:1 compatibility with the pre-trained models.
```
jmichelp, line 78: Maybe we could list the differences that could impact the traces. On top of my head we have:

- use of Chipwhisperer firmware stack (vs. bare-metal dedicated firmware in GPAM paper). Impact should be negligible when trigger is high but could have an impact (i.e. more or less delay) between the triggers
- K82F target chip clocked by Chipwhisperer at 7.37MHz vs. 10MHz or 12MHz by external oscillator in our firmware (although our paper says 7.37MHz crystal clock so I might be wrong here)
- serial communication at 38400 bauds with Chipwhisperer firmware vs. 115200 bauds in our firmware

kralka: Good point

PR #379, opened 2025-05-09
Enable choosing PicoScope by serial number

wsxrdv:


scaaml/capture/scope/ps6424e.py
```diff
@@ -427,6 +427,65 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+
+        Args:
+
+          requested_info (str): What information about the oscilloscopes is
+          requested. Defaults to empty string -- just the serial number.
+          Possible info (colon separated).
+              -v: model number
+              -c: calibration date
+              -h: hardware version
+              -u: USB version
```
kralka, line 443: No need to connect to find out what USB version is used :-)

jmichelp (approved):

PR #547, opened 2026-07-28
Pin workflows

wsxrdv:


.github/workflows/pytest.yml
jmichelp: Could you update this one too?
commit hash: `8d6379e14d29928660c4ba802d8e85393440b329` corresponds to the latest v2 (v2.3.8)

kralka: done in https://github.com/google/scaaml/pull/547/changes/6f17b49cd35d4d47b528050f9f61355817bd9bfa

kralka (approved):

PR #551, opened 2026-08-05
Implement cooldown for dependabot

jmichelp:


kralka (approved):

kralka:
The workflow checker was producing warnings, some day those might turn into errors...

